PR #23, opened 2014-08-13
Enhance sample_tools.py to allow init using local discovery doc file

jaybeeo:
Enhance discovery.py's init() function to support using a local discovery document. This is useful when the discovery document is not accessible via URL (e.g. for non-public APIs).


pferate:
This change was implemented in 63caac01ea7b9fbd8ae70bcbb4c6ecbec833ad3d.  This issue should be closed now.

nathanielmanistaatgoogle:
I agree; closing.

PR #44, opened 2015-01-07
Support Python3 for discovery module

Parkayun:
Hi.
I start supporting Python3.
First of all, discovery module is first. And I'll keep supporting another modules.

Thanks.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -344,11 +344,9 @@ def _media_size_to_long(maxSize):
   """
   if len(maxSize) < 2:
     try:
-      maxSize = eval('0L')
+      return eval('0L')
     except SyntaxError:
-      maxSize = 0
-    finally:
-      return maxSize
+      return 0
   units = maxSize[-2:].upper()
   bit_shift = _MEDIA_SIZE_BIT_SHIFTS.get(units)
   if bit_shift is not None:
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -26,17 +26,7 @@ __all__ = [
     ]
 
 
-# Support Python3
-import sys
-if sys.version_info.major > 2:
-  basestring = str
-
 # Standard library imports
-try:
-  from io import StringIO
-except ImportError:
-  import StringIO
-
 import copy
 from email.generator import Generator
 from email.mime.multipart import MIMEMultipart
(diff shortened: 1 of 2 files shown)


googleapiclient/discovery.py
```diff
@@ -757,7 +774,7 @@ def method(self, **kwargs):
     docs.append('Args:\n')
 
   # Skip undocumented params and params common to all methods.
-  skip_parameters = rootDesc.get('parameters', {}).keys()
+  skip_parameters = list(rootDesc.get('parameters', {}).keys())
```
methane: This should be `list(rootDesc.get('parameters', {}))`.

googleapiclient/discovery.py
```diff
@@ -329,7 +343,10 @@ def _media_size_to_long(maxSize):
     The size as an integer value.
   """
   if len(maxSize) < 2:
-    return 0L
+    try:
+      return eval('0L')
+    except SyntaxError:
+      return 0
```
methane: Just `return 0`

googleapiclient/discovery.py
```diff
@@ -253,8 +262,13 @@ def build_from_document(
...
   future = {}
 
-  if isinstance(service, basestring):
-    service = json.loads(service)
+  try:
+    if isinstance(service, basestring):
+      service = json.loads(service)
+  except NameError:
+    if isinstance(service, str):
+      service = json.loads(service)
```
methane: I don't like this try~except clause.
How about adding this code on top of this module?

```python
try:
    basestring
except NameError:
    basestring = str
```

Parkayun: great idea. but I change like this 
```Python
# Support Python3
import sys
if sys.version_info.major > 2:
  basestring = str
```

googleapiclient/discovery.py
Parkayun: Thanks.

googleapiclient/discovery.py
Parkayun: Any have some reason for using long literal(return 0L)? @craigcitro

craigcitro:
i think i'd prefer to use `six` rather than play games with imports. (this is the recommended approach from what i saw in the py3 porting guide?)

googleapiclient/discovery.py
craigcitro: i suspect the goal was having a consistent type on the return value.

Parkayun:
@craigcitro Oh I forgot the six. I'll commit using six.

Parkayun:
@craigcitro I fixed using six.

nathanielmanistaatgoogle:
I think that this pull request may no longer be necessary as of cf939a33f6b8 - please confirm and take action?

PR #48, opened 2015-01-20
Don't include reason in default HttpMock headers

cgurnik:
As per the [docs](https://developers.google.com/api-client-library/python/guide/mocks), the `MockHttp` object can be created as follows:

`http = HttpMock('books-discovery.json', {'status': '200'})`

The default headers for `HttpMock` currently returns a string, which causes `httplib2` to throw an exception:

`ValueError: invalid literal for int() with base 10: '200 OK'`

The relevant line from `httplib2` can be found here: https://github.com/jcgregorio/httplib2/blob/master/python2/httplib2/__init__.py#L1687

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1452,7 +1452,7 @@ class HttpMock(object):
       headers: dict, header to return with response
     """
     if headers is None:
-      headers = {'status': '200 OK'}
+      headers = {'status': '200'}
     if filename:
       f = file(filename, 'r')
       self.data = f.read()
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -999,6 +999,13 @@ class TestResponseCallback(unittest.TestCase):
     self.assertEqual(1, len(responses))
 
 
+class TestHttpMock(unittest.TestCase):
+  def test_default_response_headers(self):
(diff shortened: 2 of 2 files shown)


cgurnik:
Any interest in merging this, or should I close this?

nathanielmanistaatgoogle:
Don't close just yet. What hope is there for test coverage for this?

cgurnik:
@nathanielmanistaatgoogle Added test. Looks like Travis failed because of a GitHub issue if you wouldn't mind restarting it.

pferate:
Looks good to me for the most part.
A couple minor nitpicks:
1. Keep the double space at the end
2. The other tests use `resp.status` instead of `resp['status']`.  I would change it for consistency.

nathanielmanistaatgoogle:
Kicked the build.

Please squash commits? There's no reason for this change and its test to be in different commits.

cgurnik:
@nathanielmanistaatgoogle @pferate Comments addressed and commits squashed.

nathanielmanistaatgoogle:
I'm satisfied; will merge when @pferate also approves.

pferate:
LGTM

nathanielmanistaatgoogle:
Thanks for the contribution!

PR #53, opened 2015-02-18
Remove a newly-added exception.

craigcitro:
It turns out that the same code in `http.py` is used for processing both (1)
the initial response and redirect and (2) the replies containing actual bytes
for download. This means that the new exception we added for (2) leads to
problems in (1).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+python: 2.7
+sudo: false
+env:
+  matrix:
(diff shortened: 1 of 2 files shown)


craigcitro:
PTAL @jacktasia and @nathanielmanistaatgoogle

jacktasia:
@craigcitro  Ah, gotcha. FWIW, this still fixes the issue I originally brought up, so it seems fine to me.

PR #45, opened 2015-01-08
Python3

pferate:
I've done some work over the past few weeks and updated the entire package to work in Python 3 (3.3 and 3.4) as well as Python 2 (2.6 and 2.7).

I focused mainly on getting all of the tests to pass.  There can still be work done to clean up some parts (deprecated imports, unnecessary portions of code, etc.).  I also added a Travis-CI config and have the python3 branch passing all tests.

There was one test (test_discovery: test_methods_with_reserved_names) that would not work in both versions of python because the print statement being removed.  Maybe we can choose a different reserved name or have to separate tests.  I wasn't sure which way you want to go, so I have it skipping it in Python 3 for now.

I also took out a questionable test (test_http: test_media_io_base_upload_from_string_io).  I'm not sure if we really want to allow uploading a binary file as a string.  This works in Python 2, but is fixed and not allowed in Python 3.

PR #25, opened 2014-08-20
Prepare py3

methane:
Safe changes via 2to3 and futurize.

PR #61, opened 2015-03-03
Now uses python2.6+ style 'except'.

dkkline:
This is one teeny-tiny step towards python3 support.

Since python2.5 is not supported, and python3 is going to be supported, why not start using modern-style 'except'?

Super simple and safe step towards py3.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -205,7 +205,7 @@ def build(serviceName,
 
   try:
     service = json.loads(content)
-  except ValueError, e:
+  except ValueError as e:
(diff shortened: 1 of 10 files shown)


nathanielmanistaatgoogle:
Thanks for the contribution! This may be obviated by #25 or #45 depending on how those go - I hope we'll make some big progress in the next several hours or days.

nathanielmanistaatgoogle:
#25 just went in; please rebase/reform this pull request on top of the new head of master.

nathanielmanistaatgoogle:
I suspect that this pull request has been obviated by the work done in #64.

PR #64, opened 2015-03-04
Python3 module updates

pferate:
Got the code to pass all of the tests in python 2.6, 2.7, 3.3, and 3.4.

Removed 1 questionable test and having another test bypassed if in python 3. (see 2 of the later commits for more info)

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -28,9 +28,9 @@ __all__ = [
     'key2param',
     ]
 
+from six import StringIO
 
 # Standard library imports
-import StringIO
 import copy
 from email.generator import Generator
 from email.mime.multipart import MIMEMultipart
@@ -752,7 +752,7 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
           msgRoot.attach(msg)
           # encode the body: note that we can't use `as_string`, because
           # it plays games with `From ` lines.
-          fp = StringIO.StringIO()
+          fp = StringIO()
           g = Generator(fp, mangle_from_=False)
           g.flatten(msgRoot, unixfrom=False)
           body = fp.getvalue()
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -29,6 +29,8 @@ __all__ = [
     ]
 
 from six import StringIO
+from six.moves.urllib.parse import urlencode, urlparse, urljoin, \
+  urlunparse, parse_qsl
 
 # Standard library imports
 import copy
@@ -41,13 +43,6 @@ import logging
 import mimetypes
 import os
 import re
-import urllib
-import urlparse
-
-try:
-  from urlparse import parse_qsl
-except ImportError:
-  from cgi import parse_qsl
 
 # Third-party imports
 import httplib2
@@ -258,7 +253,7 @@ def build_from_document(
 
   if isinstance(service, six.string_types):
     service = json.loads(service)
-  base = urlparse.urljoin(service['rootUrl'], service['servicePath'])
+  base = urljoin(service['rootUrl'], service['servicePath'])
   schema = Schemas(service)
 
   if credentials:
@@ -505,7 +500,7 @@ def _urljoin(base, url):
   # exception here is the case of media uploads, where url will be an
   # absolute url.
   if url.startswith('http://') or url.startswith('https://'):
-    return urlparse.urljoin(base, url)
+    return urljoin(base, url)
   new_base = base if base.endswith('/') else base + '/'
   new_url = url[1:] if url.startswith('/') else url
   return new_base + new_url
@@ -859,14 +854,14 @@ Returns:
     request = copy.copy(previous_request)
 
     pageToken = previous_response['nextPageToken']
-    parsed = list(urlparse.urlparse(request.uri))
+    parsed = list(urlparse(request.uri))
     q = parse_qsl(parsed[4])
 
     # Find and remove old 'pageToken' value from URI
     newq = [(key, value) for (key, value) in q if key != 'pageToken']
     newq.append(('pageToken', pageToken))
-    parsed[4] = urllib.urlencode(newq)
-    uri = urlparse.urlunparse(parsed)
+    parsed[4] = urlencode(newq)
+    uri = urlunparse(parsed)
 
     request.uri = uri
 
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -46,10 +46,10 @@ import re
 
 # Third-party imports
 import httplib2
-from . import mimeparse
 import uritemplate
 
 # Local imports
+from googleapiclient import mimeparse
 from googleapiclient.errors import HttpError
 from googleapiclient.errors import InvalidJsonError
 from googleapiclient.errors import MediaUploadSizeError
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -202,6 +202,11 @@ def build(serviceName,
     raise HttpError(resp, content, uri=requested_url)
 
   try:
+    content = content.decode('utf-8')
+  except AttributeError:
+    pass
+
+  try:
     service = json.loads(content)
   except ValueError as e:
     logger.error('Failed to parse as JSON: ' + content)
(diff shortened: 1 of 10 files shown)


googleapiclient/discovery.py
```diff
@@ -712,7 +707,7 @@ def method(self, **kwargs):
         raise TypeError('media_filename must be str or MediaUpload.')
 
       # Check the maxSize
-      if maxSize > 0 and media_upload.size() > maxSize:
+      if media_upload.size() and media_upload.size() > maxSize > 0:
```
methane: Why this change?

googleapiclient/http.py
```diff
@@ -468,7 +469,7 @@ def __init__(self, body, mimetype='application/octet-stream',
     resumable: bool, True if this is a resumable upload. False means upload
       in a single request.
     """
-    fd = StringIO.StringIO(body)
+    fd = BytesIO(body)
```
methane: I think `io.BytesIO()` is OK for > Python 2.6.
`six.BytesIO()` is for <= Python 2.5.

googleapiclient/http.py
```diff
@@ -24,34 +24,35 @@
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 
-import StringIO
+from six import BytesIO, StringIO
+from six.moves.urllib.parse import urlparse, urlunparse, quote, unquote
```
methane: I feel `quote` and `unquote` are too short name.
How about `quote as urlquote, unquote as urlunquote` ?

googleapiclient/discovery.py
pferate: `media_upload.size()` isn't always defined (at least in one of the tests).  `None > 0`  will fail in python 3.

googleapiclient/http.py
pferate: I figured that I would keep it consistent and use `six` across the board for the IO's.  Is there a benefit for using `io` over `six`?

googleapiclient/http.py
pferate: It's all personal preference I guess.  I don't mind using that, but I'm not sure if there is anything in the coding standard that addresses this.

googleapiclient/discovery.py
methane: I see. I prefer `is not None` to describe `.size()` may return None.

googleapiclient/http.py
methane: Both of `six.BytesIO` and `six.StringIO` are `cStringIO.StringIO` on Python 2.
I prefer `io` for more strict behavior. But current code is OK, too.

methane:
LGTM. Great Job!

googleapiclient/discovery.py
pferate: Thinking about it a bit more, I'm thinking that it's better to check for None explicitly instead of just an object that evaluates to False.

Here is the doc for `.size()`

    Returns:
      Size of the body, or None of the size is unknown.

Added commit 8b8af28 for this.

googleapiclient/http.py
pferate: I tried looking for this in Google's style guide, but have been unable to find it anymore.  I remember reading it before, and I recall, in general `Follow PEP8, except use 2 spaces instead of 4`, but that's it.

Do you know where to find this?  We should probably add a link to it in the README.

Maybe @craigcitro or @nathanielmanistaatgoogle can help out?

pferate: I think I found what I was looking for.  I overlooked it because I thought it was more in depth.

https://code.google.com/p/google-api-python-client/wiki/BecomingAContributor

    Submitting Patches
    ...
    3. Ensure that your code adheres to the PEP 8 style, with the exception of spacing, which should be two spaces.
    ...

nathanielmanistaatgoogle:
Oh, and in case my perspective in this code review comes across as weird and unfamiliar in some respects, it might be because I make a point of reviewing provenance (https://code.google.com/p/soc/wiki/ContributingCode#Provenance) as well as content and metadata. It's certainly what's behind my asking pull requests to be of "short" (only a few commits) branches, and recently branched off head-of-master.

Note how this pull request now stands at 18 commits, some of the later of which undo the changes done in earlier commits. That's the kind of mess that should get wiped away in a commit-squashing process prior to merge of a pull request.

pferate:
Ok, thanks!  I'm going through and cleaning up the commits now.  Some of those commits that undo previous commits we put in after the initial PR was created, but they are getting cleaned up now.

pferate:
Alright.... I think we're good now.  I've taken out the extra commits and cut it down to 11 commits.  I had some fun with the rebasing, but I think I figured it out.

pferate:
It looks like there is another issue that needs to be addressed.  I think when I was developing on my Windows computer last night, the permissions got changed on a number of files and I didn't notice.  It added the execute bit, which causes nose to skip the tests.  I verified that all of the tests still pass.

From the nose documentation:
> **It is important to note that the default behavior of nose is to not include tests from files which are executable.** To include tests from such files, remove their executable bit or use the –exe flag (see ‘Options’ section below).

Let me know if you want me to fix the permissions on this PR or update the tox.ini in another PR.

nathanielmanistaatgoogle:
Fix the permissions in this pull request, please. Files in source control should not have execute bits set accidentally; they should only have them set if it is the authorial intent of the project maintainers.

pferate:
So it looks like `@unittest.skip()` and `@unittest.skipif()` were added in 2.7, so the 2 tests fail as is in 2.6 (see the travis logs).  It looks like we can use the package `unittest2`.
> unittest2 is a backport of the new features added to the unittest testing framework in Python 2.7 and onwards. It is tested to run on Python 2.6, 2.7, 3.2, 3.3, 3.4 and pypy.

I know that it isn't specific to Python3, but the `skipif()` was introduced in this PR.  There are a few ways to handle this, but how would you like it done?

nathanielmanistaatgoogle:
I think that what you've done in fde7488 looks like the right way to use unittest2 to handle 2.6 compatibility.

Looking forward: fbae355b05 and fde7488 shouldn't exist as they merely tweak work done earlier on the not-yet-merged-to-master line of development. I think you should simply amend those earlier commits to fix the issues in them.

Also on the matter of amending commits: please follow the guidelines at chris.beams.io/posts/git-commit/#seven-rules. We should be able to look at https://github.com/google/google-api-python-client/pull/64/commits and not see any subject lines cut off and prematurely ending in ellipses.

This feels like it's getting close! :-)

pferate:
I've learned a lot about rewriting history these past few days. :)

fbae355 has been incorporated to previous commits. (It was interesting splitting it up and squashing to previous commits)

fde7488 was needed for the 2 tests that were modified in 749477c, but I did it for all files for consistancy.  In my mind they could be seen as atomic and in separate commits, but maybe in reverse order.

I'll modify the commit messages and move fde7488 until I hear otherwise.

nathanielmanistaatgoogle:
Sounds like the right approach.

One more minor nitpick: "consistency", not "consistancy".

pferate:
Alright... I think I got it all.  I broke the history for a bit, but I think I got everything back in it's place.

At least I was spelling it "consistancy" with consistency... :-p

pferate:
@nathanielmanistaatgoogle: Alright, I've made the last few changes, how does it look now?

nathanielmanistaatgoogle:
Looks pretty good, but where is the separate pull request introducing unittest2? It may have been formed out of this but it should be merged and then this pull request should be rebased on top of it.

pferate:
I created #66 to introduce unittest2.  Once it gets merged, I'll rebase this branch to the merge commit.

pferate:
Rebased this branch to 4e7e6d40270ce0f46cdd54e2c26eee73a233d11a.

nathanielmanistaatgoogle:
Made it the whole way through, finally, so (at least in theory) I'm done asking questions about the change. :-)

pferate:
I updated commit ca397a7 to just check the current version of python and run the appropriate check.

nathanielmanistaatgoogle:
I think this is ready for merge.

@methane may I ask you to make another quick re-review since a few things have changed over the course of the conversation?

methane:
LGTM

PR #67, opened 2015-03-10
Including changes for Python3 in setup.py

pferate:
Updating setup.py file to include Python3.

1. I used the latest version (1.4.6) of oauth2client, although 1.4 was the version the introduced Py3 support.
2. I changed the Operating System to OS Independent since it is.  I'm not sure if I should put this in a separate PR or commit since it's not needed for Py3, but it's only one line.
3. I'll leave updating the version number and documentation up to you Googlers since you will be the ones pushing this out.


nathanielmanistaatgoogle:
Thanks!

PR #73, opened 2015-03-19
Removed unneeded/unused line of code

mmascolino:
The call to build(api, version, http) to setup the service stub is unneeded since the call to sample_tools.init above does that for the sample.  Plus the build function was not imported from apiclient.discovery which causes this sample to not work at all.


mmascolino:
I signed the CLA.

nathanielmanistaatgoogle:
Thanks for the contribution!

PR #76, opened 2015-03-24
py3 errors: decode `bytes` input before passing it to json.loads

saaros:
Http response contents use `bytes` type in Python 3.  json.loads requires
its input to be `str` so utf-8 decode the input before trying to load it.

### DIFF
--- a/googleapiclient/errors.py
+++ b/googleapiclient/errors.py
@@ -44,7 +44,12 @@ class HttpError(Error):
     """Calculate the reason for the error from the response content."""
     reason = self.resp.reason
     try:
-      data = json.loads(self.content)
+      # turn content into a unicode string, this is required on python 3
+      if isinstance(self.content, bytes):
+        content = self.content.decode('utf-8')
+      else:
+        content = self.content
+      data = json.loads(content)
       reason = data['error']['message']
     except (ValueError, KeyError):
       pass
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -71,6 +71,17 @@ class Error(unittest.TestCase):
     error = HttpError(resp, content)
     self.assertEqual(str(error), '<HttpError 400 "Failed">')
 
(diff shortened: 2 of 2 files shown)


nathanielmanistaatgoogle:
@saaros: what hope is there for test coverage for this change?

@pferate and @methane: what do the two of you think of this change?

saaros:
@nathanielmanistaatgoogle added a simple test case and pushed an update.

methane:
I think `fake_response()` should return content in `bytes` like httplib2.

methane:
HTTPMockSequence should return content as bytes, too.

googleapiclient/errors.py
```diff
@@ -44,7 +44,13 @@ def _get_reason(self):
     """Calculate the reason for the error from the response content."""
     reason = self.resp.reason
     try:
-      data = json.loads(self.content)
+      # json.loads can't handle bytes inputs on python 3, on python 2 bytes
+      # is the same as str
+      if isinstance(self.content, bytes) and bytes is not str:
```
methane: I feel `not six.PY2` is clearer than `bytes is not str`.

pferate:
I think that instead of testing if it's bytes or strings first, put it in a try: ... except: block.  Similar to what was done in 9b0452c7683da683372821861e7bcf06036b2832.

methane:
I've made #77.

methane:
@pferate I doesn't like it.
I think `bytes or unicode` everywhere is anti-pattern when porting PY2 code to PY2/3 code.

saaros:
Dropped the pointless `bytes is not str` check, let's just utf8-decode the contents if they're bytes.

Feel free to close this PR if you prefer the approach in #77.

saaros:
Also, I don't think the `try: s.decode("utf-8"); except AttributeError: pass` thing is a good pattern, it's slow and will break if you happen to pass it a unicode input on Python 2.

PR #77, opened 2015-03-24
Fix unicode issues in HttpError and BatchHttpRequest

methane:
1) HttpError parses http response by json.loads. It accepts only
unicode on PY3 while response is bytes. So decode response first.

2) BatchHttpRequest uses email.parser.FeedParser. It accepts only
unicode on PY3, too. So encode parsed response to emulate normal
http response.

### DIFF
--- a/googleapiclient/errors.py
+++ b/googleapiclient/errors.py
@@ -37,6 +37,8 @@ class HttpError(Error):
   @util.positional(3)
   def __init__(self, resp, content, uri=None):
     self.resp = resp
+    if not isinstance(content, bytes):
+        raise TypeError("HTTP content should be bytes")
     self.content = content
     self.uri = uri
 
@@ -44,7 +46,7 @@ class HttpError(Error):
     """Calculate the reason for the error from the response content."""
     reason = self.resp.reason
     try:
-      data = json.loads(self.content)
+      data = json.loads(self.content.decode('utf-8'))
       reason = data['error']['message']
     except (ValueError, KeyError):
       pass
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1257,6 +1257,10 @@ class BatchHttpRequest(object):
 
     # Prepend with a content-type header so FeedParser can handle it.
     header = 'content-type: %s\r\n\r\n' % resp['content-type']
+    # PY3's FeedParser only accepts unicode. So we should decode content
+    # here, and encode each payload again.
+    if six.PY3:
+      content = content.decode('utf-8')
     for_parser = header + content
 
     parser = FeedParser()
@@ -1270,6 +1274,9 @@ class BatchHttpRequest(object):
     for part in mime_response.get_payload():
       request_id = self._header_to_id(part['Content-ID'])
       response, content = self._deserialize_response(part.get_payload())
+      # We encode content here to emulate normal http response.
+      if isinstance(content, six.text_type):
+        content = content.encode('utf-8')
       self._responses[request_id] = (response, content)
 
   @util.positional(1)
@@ -1536,6 +1543,8 @@ class HttpMockSequence(object):
         content = body
     elif content == 'echo_request_uri':
       content = uri
+    if isinstance(content, six.text_type):
+      content = content.encode('utf-8')
     return httplib2.Response(resp), content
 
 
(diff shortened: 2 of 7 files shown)


nathanielmanistaatgoogle:
(1) Squash commits?
(2) Write a longer, more detailed commit message? (http://chris.beams.io/posts/git-commit/#seven-rules)
(3) Why does Travis CI show test failure for this change?

googleapiclient/errors.py
```diff
@@ -44,9 +44,9 @@ def _get_reason(self):
     """Calculate the reason for the error from the response content."""
     reason = self.resp.reason
     try:
-      data = json.loads(self.content)
+      data = json.loads(self.content.decode('utf-8'))
```
pferate: I think that this should not be included within this block, it should be in a separate one before this.

If there is actually a 'reason' but it fails on UnicodeError, it will seem as if there is no reason when in fact there is.

googleapiclient/errors.py
```diff
@@ -44,9 +44,9 @@ def _get_reason(self):
...
       reason = data['error']['message']
-    except (ValueError, KeyError):
+    except (ValueError, KeyError, UnicodeError):
```
pferate: It should be AttributeError (See output of nosetests/Travis)

pferate:
@nathanielmanistaatgoogle 
1) I was going to say the same, I've fought that battle a few times here recently
2) I like that post, what do you think about linking it somewhere in the README?
3) The wrong Error was being caught, I pointed it out in the line notes.

googleapiclient/http.py
```diff
@@ -1252,11 +1252,10 @@ def _execute(self, http, order, requests):
     if resp.status >= 300:
       raise HttpError(resp, content, uri=self._batch_uri)
 
-    # Now break out the individual responses and store each one.
```
pferate: I agree that this shouldn't be here, but it is not related to the rest of the commit/PR; move it to a separate commit/PR.

pferate:
There are 3 ways that you have for getting to the decode/encode calls:
1) try/except
2) if not six.PY2:
3) if isinstance(content, six.text_type):

Personally, I like "It's better to ask forgiveness than permission", so I would go with number 1.  I'm not sure if @nathanielmanistaatgoogle has a preference.

googleapiclient/errors.py
methane: http response should bytes always. It should have `.decode()` on both of PY2 and PY3.
AttributeError is just a bug. Catching it is bad idea.

googleapiclient/http.py
methane: done

methane:
> (1) Squash commits?
Done

> (2) Write a longer, more detailed commit message? (http://chris.beams.io/posts/git-commit/#seven-rules)
Done, bug I'm not a good English writer.

> (3) Why does Travis CI show test failure for this change?
There were bugs remained relating BatchHttpRequest. I've fixed it.

googleapiclient/http.py
```diff
@@ -1257,6 +1257,8 @@ def _execute(self, http, order, requests):
 
     # Prepend with a content-type header so FeedParser can handle it.
     header = 'content-type: %s\r\n\r\n' % resp['content-type']
+    if not six.PY2:
+      content = content.decode('utf-8')
```
methane: May content have binary data?
If so, [surrogate escape](https://www.python.org/dev/peps/pep-0383/) may help us.
But I'm not sure about how FeedParser treat surrogate escaped text.

googleapiclient/errors.py
methane: I've added assert in `__init__` to catch bugs early.

pferate: Looking at it again, I think your idea is better.  What do you think about checking for bytes and raising an error instead of using the assert?

methane: > Looking at it again, I think your idea is better. What do you think about checking for bytes and raising an error instead of using the assert?

I've did it. Raise TypeError instead of AssertionError

googleapiclient/errors.py
```diff
@@ -37,6 +37,9 @@ class HttpError(Error):
   @util.positional(3)
   def __init__(self, resp, content, uri=None):
     self.resp = resp
+    if not isinstance(content, bytes):
+        # When see this exception in test, testdata may be wrong
```
nathanielmanistaatgoogle: Drop this comment; test data passing through this function should be all corrected in this commit.

methane: done

pferate:
LGTM.

nathanielmanistaatgoogle:
Thanks for the contribution. :-)

PR #79, opened 2015-03-27
Adding coveralls integration

pferate:
Run coverage with all tests and send results to coveralls.io.

This was added to oauth2client and I figured we should add it here as well.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,5 +9,6 @@ deps = keyring
        django
        webtest
        nose
+       coverage>=3.6,<3.99
        unittest2
-commands = nosetests
+commands = nosetests --with-coverage --cover-package=googleapiclient --nocapture --cover-erase --cover-tests --cover-branches --cover-min-percentage=85
--- a/tox.ini
+++ b/tox.ini
@@ -10,16 +10,4 @@ deps = keyring
        webtest
        nose
        unittest2
-setenv = PYTHONPATH=../google_appengine
-
-[testenv:py26]
-commands = nosetests --ignore-files=test_oauth2client_appengine\.py
-
-[testenv:py27]
-commands = nosetests
-
-[testenv:py33]
-commands = nosetests
-
-[testenv:py34]
 commands = nosetests
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[report]
+omit = */samples/*
+exclude_lines =
+    # Re-enable the standard pragma
+    pragma: NO COVER
+    # Ignore debug-only repr
+    def __repr__
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@ dist/
 
 # Test files
 .tox/
+
+# Coverage files
+.coverage
+coverage.xml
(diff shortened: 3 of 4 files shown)


tox.ini
```diff
@@ -9,17 +9,18 @@ deps = keyring
...
        unittest2
 setenv = PYTHONPATH=../google_appengine
 
 [testenv:py26]
-commands = nosetests --ignore-files=test_oauth2client_appengine\.py
+commands = nosetests --with-coverage --cover-package=googleapiclient --nocapture --cover-erase --cover-tests --cover-branches --cover-min-percentage=85
 
 [testenv:py27]
-commands = nosetests
+commands = {[testenv:py26]commands}
```
nathanielmanistaatgoogle: What's the significance of "testenv:py26" on lines twenty, twenty-three, and twenty-six? Should those lines not be "py27", "py33", and "py34"?

nathanielmanistaatgoogle:
When I look at oauth2client's coveralls setup, I see "cover" as its own tox environment. Why is that not the right thing to do here?

tox.ini
pferate: It's just referencing the commands in from py26 to make ensure the same syntax is used throughout.

Actually, I just tested something out locally and found a way to may it a whole lot cleaner.  I'll update the PR soon.

pferate:
We will probably want to fix that in oauth2client.  Right now, it only checks the coverage in 2.7.  The way it is implemented there, the nose tests are ran normally first, then in 2.7, the nose tests are ran again with the coverage check; so the test are running twice when only once is needed (per version).

I don't see a reason to *not* check the coverage for all versions when the tests are ran the first time, since some of the code only runs in 2.x and others only in 3.x.  Coveralls will also give you a summary for all of the test environments afterwards.

tox.ini
```diff
@@ -9,17 +9,7 @@ deps = keyring
        django
        webtest
        nose
+       coverage>=3.6,<3.99
        unittest2
 setenv = PYTHONPATH=../google_appengine
-
-[testenv:py26]
-commands = nosetests --ignore-files=test_oauth2client_appengine\.py
```
nathanielmanistaatgoogle: Why is the loss of this --ignore-files bit okay? It looks important, it smells important, and it's completely unmentioned in the commit log message?

pferate: It looks like the file got removed sometime between 2013 and now.  I'm trying to look through the git logs to find when it got removed.

pferate: Commit 51852247883c780b68006d6d6b4888b6efbdf6c7 deleted all of the oauth2client tests since they were included in oauth2client.

nathanielmanistaatgoogle: Please split the commit. "Simplifying the tox file as a result of months-in-the-past changes" is significantly different enough from "add coveralls integration" that it's the sensible thing to spread them across two commits.

Thanks!

pferate:
Split the commit into 2, how's it look now?

nathanielmanistaatgoogle:
Looks much better; thanks!

PR #81, opened 2015-03-31
Python 3 compat improvements

saaros:
* tests: use a different name for reserved name test and enable on py3
* json input: don't try to utf-8 decode everything unconditionally

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -201,10 +201,8 @@ def build(serviceName,
   if resp.status >= 400:
     raise HttpError(resp, content, uri=requested_url)
 
-  try:
+  if isinstance(content, bytes):
     content = content.decode('utf-8')
-  except AttributeError:
-    pass
 
   try:
     service = json.loads(content)
--- a/googleapiclient/model.py
+++ b/googleapiclient/model.py
@@ -258,10 +258,8 @@ class JsonModel(BaseModel):
     return json.dumps(body_value)
 
   def deserialize(self, content):
-    try:
-        content = content.decode('utf-8')
-    except AttributeError:
-        pass
+    if isinstance(content, bytes):
+      content = content.decode('utf-8')
     body = json.loads(content)
     if self._data_wrapper and isinstance(body, dict) and 'data' in body:
       body = body['data']
--- a/tests/data/zoo.json
+++ b/tests/data/zoo.json
@@ -288,10 +288,10 @@
   },
   "global": {
    "resources": {
-    "print": {
+    "from": {
      "methods": {
       "assert": {
-       "path": "global/print/assert",
+       "path": "global/from/assert",
        "id": "zoo.animals.mine",
        "httpMethod": "GET",
        "parameters": {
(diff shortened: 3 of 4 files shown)


nathanielmanistaatgoogle:
Looks good but I'd like to have another reviewer also approve. @pferate, @methane, what do you think?

googleapiclient/discovery.py
```diff
@@ -201,10 +201,8 @@ def build(serviceName,
   if resp.status >= 400:
     raise HttpError(resp, content, uri=requested_url)
 
-  try:
+  if isinstance(content, bytes):
     content = content.decode('utf-8')
-  except AttributeError:
-    pass
```
methane: I think `http.request` always return contents as `bytes`.
So we can just remove this `try`~`except`.
Or `assert isinstance(content, bytes)`.

googleapiclient/model.py
```diff
@@ -258,10 +258,8 @@ def serialize(self, body_value):
     return json.dumps(body_value)
 
   def deserialize(self, content):
-    try:
-        content = content.decode('utf-8')
-    except AttributeError:
-        pass
+    if isinstance(content, bytes):
+      content = content.decode('utf-8')
```
methane: docstring in `BaseModel.deserialize()` says type of content is string.
I think this project (or google) should decide what 'string' means.

My proposal is:

* string -- basestring in PY2, str in PY3
* bytes -- str(=bytes) in PY2, bytes in PY3
* unicode -- unicode in PY2, str in PY3

googleapiclient/discovery.py
saaros: Well - theoretically another http client (or proxy) could encode response in a different character set and to pass the data properly to this code the application would have to decode the bytes into unicode and the encode them in utf-8 before passing them to this function which would again decode it.  I think it's better to allow unicode objects (str or unicode) and try to handle all bytes inputs as utf-8.

googleapiclient/model.py
saaros: I think a string should mean a textual string, basically unicode in py2 and str in py3.  And for compatibility all bytes inputs should be handled as utf-8 encoded strings.

googleapiclient/discovery.py
pferate: I agree with @methane here.  Although the flexibility of handling both is nice (and how I initially handled it), httplib2 states:
> In httplib2, the response headers are strings, but the content is bytes.

From some of the docstrings in this library, 
> http: httplib2.Http, An instance of httplib2.Http or something that acts like it that HTTP requests will be made through.

So if another http client is used here, it should follow the same rules.  We should also not let those rules be broken here as well.

saaros: You can change it into an assert if you think it makes more sense.  I don't think it does, though, as you're anyway turning the input into a unicode string and could just as well accept an already-unicode input.  But in any case I'd like to get rid of the current, broken, code.

jonparrott:
@saaros I'm closing this PR due to inactivity. If you're still interested in getting this merged please let us know and we'll be happy to re-open it.

PR #82, opened 2015-04-01
be completed if the HTTP status 200 is responded for range request.

crkz:
The current implementation is checking for transferred progress of total size to completion judgment.
will prevent an infinite loop of the case that are not contain both "Content-Range" and "Content-Length" is  in the response.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -566,7 +566,7 @@ class MediaIoBaseDownload(object):
       elif 'content-length' in resp:
         self._total_size = int(resp['content-length'])
 
-      if self._progress == self._total_size:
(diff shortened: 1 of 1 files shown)


crkz:
CLA signed.

nathanielmanistaatgoogle:
Why was this closed?

crkz:
I've signed with another email address... I made another p-r with appropriate one.

nathanielmanistaatgoogle:
Understood. In the future, there's no need for that, because that sort of problem can be corrected inside an open pull request. :-)

crkz:
Thanks for your advice, noted!

PR #83, opened 2015-04-01
infinite loop when range request has been ignored and no length information in response header

crkz:
I've been experienced both "Content-Length" and "Content-Range" is not in response while download from Google Cloud Storage.

Current implementation is checked by total length. but sometime less.
set "done" flag by HTTP Status 200 (not 206 Partial) is enough i guess.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -565,6 +565,8 @@ class MediaIoBaseDownload(object):
         self._total_size = int(length)
       elif 'content-length' in resp:
         self._total_size = int(resp['content-length'])
+      elif resp.status == 200:
+        self._done = True
 
       if self._progress == self._total_size:
         self._done = True
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -348,6 +348,29 @@ class TestMediaIoBaseDownload(unittest.TestCase):
     self.assertEqual(5, download._progress)
     self.assertEqual(5, download._total_size)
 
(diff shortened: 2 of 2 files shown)


nathanielmanistaatgoogle:
What's up with the test failures on Travis?

crkz:
idk about Travis CI failed...
any special instruction required to fix problem?

nathanielmanistaatgoogle:
Do the tests pass when you run them locally? Does the log of the failure available on Travis offer you any insight as to what happened?

crkz:
Let me check what is point. I'll commit after test passed locally.

nathanielmanistaatgoogle:
(1) What hope is there for test coverage for this change?

(2) Please amend your commit message to conform to chris.beams.io/posts/git-commit/#seven-rules.

crkz:
(1) set "chunksize" value as greater than total size, and no "Content-Range" and "Content-Length" header and status "200" in response will reproduce the problem I guess. May I contribute test code too? :)

(2) I'll amend after clarified above.

nathanielmanistaatgoogle:
Yes, absolutely, please add new test code that fails in the absence of your change to production code.

crkz:
Copy that. I'll write unit test then. Thanks.

jonparrott:
@crkz I'm closing this PR due to inactivity. If you're still interested in getting this merged please let us know and we'll be happy to re-open it.

PR #87, opened 2015-04-11
Fix error reporting in python 3

JelteF:
The content consists of bytes in python 3 and the HttpError object thinks its a string when it tries to decode it. I'm not sure if this is the best place to fix it, since it could also be fixed in the HttpError object itself.


JelteF:
Apperently there is already a test for something like this, but it seems to fail anyway.

methane:
Have you read #77?

JelteF:
No, that probably fixed my problem. In that case I will use the master
branch in git again.

PR #88, opened 2015-04-15
Minor Markdown Fix

mmattioli:
Proposal
===

Update the copyright years on some files to 2015. Update some .md files to use better header formatting.

Reasoning
===

The nature of this PR should be self-explanatory but, just incase, we are currently in the year 2015 and the copyright years should reflect such as this is still an active project. The header style change in the .md files preserves the intended visual but uses a more easily editable method (no one wants to keep typing === under every letter of the header).

### DIFF
--- a/samples/adexchangebuyer/README.md
+++ b/samples/adexchangebuyer/README.md
@@ -1,5 +1,5 @@
-Where are the samples?
-=======================================================
+# Where are the samples?
+
 The DoubleClick Ad Exchange Buyer REST API samples are now hosted separately
 from the client library project. You can find them
-[here.](https://github.com/googleads/googleads-adxbuyer-examples/tree/master/python)
+[here](https://github.com/googleads/googleads-adxbuyer-examples/tree/master/python).
--- a/samples/dfareporting/README.md
+++ b/samples/dfareporting/README.md
@@ -1,5 +1,5 @@
-Where are the samples?
-=======================================================
+# Where are the samples?
+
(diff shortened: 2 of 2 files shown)


nathanielmanistaatgoogle:
The markdown change looks righteous but I'll have to ask my colleagues about changing the copyright notices.

mmattioli:
:+1:

mmattioli:
@nathanielmanistaatgoogle just wondering what the status of this is. Thanks.

dberlin:
Engilawyer here.
While we appreciate you taking the time to do this, updating copyright years is actually not legally required, and doesn't legally achieve anything (it's essentially cargo cult lawyering).  The date of first publication is clearly valid, and using other dates is actually "risky" (IE adding more dates could make your copyright notice invalid, even though it's common practice).
So we'd rather avoid the churn.

It's true that some people use copyright notices as a proxy for "how out of date something is", but that'll never change unless people stop updating them pointlessly each year :)

nathanielmanistaatgoogle:
Thank you @dberlin.

@mmattioli: please reform this pull request to include just the documentation improvements and thank you for your contribution.

mmattioli:
@nathanielmanistaatgoogle I just reverted in 3b5096d488a9f01fc6e85c569a37c9f31bca3741. This PR now includes only the doc fixes

mmattioli:
@nathanielmanistaatgoogle I just committed with the documentation improvements only. Thanks!

PR #101, opened 2015-06-08
Stop stripping newlines from batch requests

eljobe:
When forming a multipart/mixed HTTP request like the ones created using
the BatchHttpRequest object, the request is supposed to include two
trailing newlines to signify:

1. The termination of the HTTP headers for each wrapped request.
2. The empty body of the wrapped request.

Without this change some implementations of API servers were rejecting
requests generated by this client.  Specifically requests which had an
empty wrapped request body.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1120,10 +1120,6 @@ class BatchHttpRequest(object):
     g.flatten(msg, unixfrom=False)
     body = fp.getvalue()
 
-    # Strip off the \n\n that the MIME lib tacks onto the end of the payload.
-    if request.body is None:
-      body = body[:-2]
-
     return status_line + body
 
   def _deserialize_response(self, payload):
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -453,6 +453,11 @@ MIME-Version: 1.0
 Host: www.googleapis.com
 content-length: 0\r\n\r\n"""
 
+NO_BODY_EXPECTED_GET = """GET /someapi/v1/collection/?foo=bar HTTP/1.1
+Content-Type: application/json
+MIME-Version: 1.0
+Host: www.googleapis.com\r\n\r\n"""
+
 
 RESPONSE = """HTTP/1.1 200 OK
 Content-Type: application/json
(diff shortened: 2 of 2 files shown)


googleapiclient/http.py
```diff
@@ -1120,10 +1120,6 @@ def _serialize_request(self, request):
     g.flatten(msg, unixfrom=False)
     body = fp.getvalue()
 
-    # Strip off the \n\n that the MIME lib tacks onto the end of the payload.
-    if request.body is None:
```
nathanielmanistaatgoogle: I don't think I like the blind truncation here - is it guaranteed by the MIME library that the string it returns will end in two newlines, or is that an implementation detail? Let's be a little safer here and say

if request.body is None and body[-2:] == '\n\n':

? Why is

body = body.rstrip()

not the appropriate action to take here?

nathanielmanistaatgoogle:
Please amend your commit message to follow the guidelines at chris.beams.io/posts/git-commit/#seven-rules.

nathanielmanistaatgoogle:
What hope is there of test coverage for this change?

googleapiclient/http.py
eljobe: I'm not sure.  But, are you reading these line deletions as additions?  This change is to remove the stripping, not to add it.

eljobe:
I've reformatted the commit message according to the guidelines above.
I'm not sure what the best way to test this is.  If you'll give me a pointer, I'm not against adding a test.

googleapiclient/http.py
nathanielmanistaatgoogle: Whoops! Must have a case of the Mondays.

nathanielmanistaatgoogle:
GitHub isn't yet showing me the reformatted commit message (the subject line is still much more than fifty characters).

Have you looked at "test_serialize_request", "test_serialize_request_media_body", and "test_serialize_request_no_body" in tests/test_http.py? Could you augment one of those methods or add another sibling method to the family?

eljobe:
Okay.  Test added, and git commit messages fixed.

nathanielmanistaatgoogle:
LGTM. @craigcitro or @soltanmm any additional comments here?

craigcitro:
LGTM

so i'm vaguely curious why this was done in the first place, since it's clearly an explicit choice. 

on the flipside, i know @eljobe has tested this, so if it works, maybe we should just roll with it.

nathanielmanistaatgoogle:
It appears to have been done three and a half years ago in 66f5752dacf3820c6f388e (reviewed at https://codereview.appspot.com/5434059/) by @jcgregorio. There doesn't appear to be any further remark about it at the time, but the comment "Strip off the \n\n that the MIME lib tacks onto the end of the payload" carries a certain deliberateness.

I guess what we should say up front is "the MIME lib is right to tack on those trailing newlines"?

eljobe:
When you say, "say up front" do you mean you want me to add a comment to the code, or more information to one of the commit messages or something?  I'm not sure how to move this change forward.

nathanielmanistaatgoogle:
I was curious to hear from @jcgregorio about the original motivation for the newline stripping but there's no reason to block on his response and honestly who among us can ever account for small oddities in years-ago-written code anyway?

Now that I reread your commit message, there's nothing wanting from it that I think needs to be added.

Merging.

PR #105, opened 2015-06-12
Add a new_batch_http_request() method to services

eljobe:
This method can be used to get a BatchHttpRequest object with the
batch_uri calculated from the rootUrl and batchPath fields from the
discovery doc.

Without this change, users have to know what the batch url for newer api
endpoints is.  For example:
  batch = BatchHttpRequest(batch_uri=http://zoo.googleapis.com/batch)

With this change, the syntax is easier:
  zoo = build('zoo', 'v1', http=http)
  batch = zoo.new_batch_http_request()

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -56,6 +56,7 @@ from googleapiclient.errors import MediaUploadSizeError
 from googleapiclient.errors import UnacceptableMimeTypeError
 from googleapiclient.errors import UnknownApiNameOrVersion
 from googleapiclient.errors import UnknownFileType
+from googleapiclient.http import BatchHttpRequest
 from googleapiclient.http import HttpRequest
 from googleapiclient.http import MediaFileUpload
 from googleapiclient.http import MediaUpload
@@ -950,6 +951,14 @@ class Resource(object):
     self._add_next_methods(self._resourceDesc, self._schema)
 
   def _add_basic_methods(self, resourceDesc, rootDesc, schema):
+    # If this is the root Resource, add a new_batch_http_request() method.
+    if resourceDesc == rootDesc:
+      batch_uri = '%s%s' % (
+        rootDesc['rootUrl'], rootDesc.get('batchPath', 'batch'))
+      def new_batch_http_request(callback=None):
+        return BatchHttpRequest(callback=callback, batch_uri=batch_uri)
+      self._set_dynamic_attr('new_batch_http_request', new_batch_http_request)
+
     # Add basic methods to Resource
     if 'methods' in resourceDesc:
       for methodName, methodDesc in six.iteritems(resourceDesc['methods']):
--- a/tests/test_discovery.py
+++ b/tests/test_discovery.py
@@ -55,6 +55,7 @@ from googleapiclient.errors import InvalidJsonError
 from googleapiclient.errors import MediaUploadSizeError
 from googleapiclient.errors import ResumableUploadError
 from googleapiclient.errors import UnacceptableMimeTypeError
+from googleapiclient.http import BatchHttpRequest
 from googleapiclient.http import HttpMock
 from googleapiclient.http import HttpMockSequence
 from googleapiclient.http import MediaFileUpload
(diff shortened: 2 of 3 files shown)


googleapiclient/discovery.py
```diff
@@ -950,6 +951,17 @@ def _set_service_methods(self):
     self._add_next_methods(self._resourceDesc, self._schema)
 
   def _add_basic_methods(self, resourceDesc, rootDesc, schema):
+    # If this is the root Resource, add a new_batch_http_request() method.
+    if resourceDesc == rootDesc:
+      def new_batch_http_request(callback=None):
+        batchPath = 'batch'
```
nathanielmanistaatgoogle: How about

batchPath = rootDesc.get('batchPath', 'batch')

instead of these three lines?

googleapiclient/discovery.py
```diff
@@ -950,6 +951,17 @@ def _set_service_methods(self):
...
+        if 'batchPath' in rootDesc:
+          batchPath = rootDesc['batchPath']
+        return BatchHttpRequest(
+          callback = callback,
```
nathanielmanistaatgoogle: No space on either side of an equals sign used to specify a keyword argument. (On this line and the next line.)

tests/test_discovery.py
```diff
@@ -514,6 +515,22 @@ def test_patch(self):
 
     self.assertEqual(request.method, 'PATCH')
 
+  def test_batch_request_from_discovery(self):
+    self.http = HttpMock(datafile('zoo.json'), {'status': '200'})
+    # zoo defines a batchPath
+    zoo = build('zoo', 'v1', http=self.http)
+    expected = "https://www.googleapis.com/batchZoo"
```
nathanielmanistaatgoogle: Drop "expected" as a local field and just use this literal value directly at the field's one use-site.

tests/test_discovery.py
```diff
@@ -514,6 +515,22 @@ def test_patch(self):
...
+    actual = zoo.new_batch_http_request()
```
nathanielmanistaatgoogle: I'm wary of identifiers that are only adjectives. This is a BatchHttpRequest, right? How about calling it "batch_request"?

tests/test_discovery.py
```diff
@@ -514,6 +515,22 @@ def test_patch(self):
...
+    self.assertEqual(actual._batch_uri, expected)
+
+  def test_batch_request_from_default(self):
+    self.http = HttpMock(datafile('plus.json'), {'status': '200'})
+    # plus does not define a batchPath
+    plus = build('plus', 'v1', http=self.http)
+    expected = "https://www.googleapis.com/batch"
```
nathanielmanistaatgoogle: Same comments in this function as in the above function.

nathanielmanistaatgoogle:
Spelling errors on both the first and last lines of your commit message, right? Plus an erroneous second space after a period in the middle paragraph (http://www.slate.com/articles/technology/technology/2011/01/space_invaders.html).

googleapiclient/discovery.py
```diff
@@ -950,6 +951,17 @@ def _set_service_methods(self):
     self._add_next_methods(self._resourceDesc, self._schema)
 
   def _add_basic_methods(self, resourceDesc, rootDesc, schema):
+    # If this is the root Resource, add a new_batch_http_request() method.
+    if resourceDesc == rootDesc:
+      def new_batch_http_request(callback=None):
+        batchPath = 'batch'
+        if 'batchPath' in rootDesc:
+          batchPath = rootDesc['batchPath']
```
nathanielmanistaatgoogle: How significant is it that batchPath is computed here instead of outside of the nested method def? Is it expected to change between successive calls to new_batch_http_request?

googleapiclient/discovery.py
```diff
@@ -950,6 +951,17 @@ def _set_service_methods(self):
...
+        return BatchHttpRequest(
+          callback = callback,
+          batch_uri = '%s%s' % (rootDesc['rootUrl'], batchPath))
```
nathanielmanistaatgoogle: Is this value expected to change between successive calls to new_batch_http_request?

nathanielmanistaatgoogle:
Given the dynamism here I'll want @craigcitro to be looking at this as well as myself.

googleapiclient/discovery.py
eljobe: Done, and moved the batch_uri construction up here as well.

googleapiclient/discovery.py
eljobe: Done.

tests/test_discovery.py
eljobe: Done.

tests/test_discovery.py
eljobe: Done.

tests/test_discovery.py
eljobe: Done.

googleapiclient/discovery.py
eljobe: Good point. It is not expected to change, I've moved the construction of the uri to outside the method definition.

googleapiclient/discovery.py
eljobe: Ditto.

googleapiclient/discovery.py
```diff
@@ -950,6 +951,14 @@ def _set_service_methods(self):
     self._add_next_methods(self._resourceDesc, self._schema)
 
   def _add_basic_methods(self, resourceDesc, rootDesc, schema):
+    # If this is the root Resource, add a new_batch_http_request() method.
+    batch_uri = '%s%s' % (
```
nathanielmanistaatgoogle: Move this calculation inside your if since its result will only be used inside your if.

eljobe: Done.

eljobe:
Took me a little while to figure out git rebase, but this now looks like a single commit like we wanted, right?

nathanielmanistaatgoogle:
LGTM. Any urgent need for this to be merged right away?

craigcitro:
LGTM

PR #106, opened 2015-06-13
Adds a doc comment for new_batch_http_request

eljobe:
This comment doesn't actually show up in the generated epydoc, but it
does allow the "make docs" command to complete, and it is nice to have
the explanation of how to use the method in the source repository.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -956,6 +956,19 @@ class Resource(object):
       batch_uri = '%s%s' % (
         rootDesc['rootUrl'], rootDesc.get('batchPath', 'batch'))
       def new_batch_http_request(callback=None):
+        """Create a BatchHttpRequest object based on the discovery document.
+
+        Args:
+          callback: callable, A callback to be called for each response, of the
+            form callback(id, response, exception). The first parameter is the
+            request id, and the second is the deserialized response object. The
+            third is an apiclient.errors.HttpError exception object if an HTTP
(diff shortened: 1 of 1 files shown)


nathanielmanistaatgoogle:
If "make docs" is currently broken without this change, and this change fixes "make docs", I'd like that described in greater detail in the commit message?

nathanielmanistaatgoogle:
Also please get the CLA signature business squared away. Is there something different about this contribution that was different about your other contribution, because we didn't see the bot object over there, did we?

eljobe:
Added information to the git commit message and fixed the authorship of the commit.

googleapiclient/discovery.py
```diff
@@ -956,6 +956,16 @@ def _add_basic_methods(self, resourceDesc, rootDesc, schema):
...
       def new_batch_http_request(callback=None):
+        """Create a BatchHttpRequest object based on the discovery document.
+
+        Args:
+          callback: callable, A callback to be called for each response, of the
+            form callback(id, response, exception). The first parameter is the
+            request id, and the second is the deserialized response object. The
+            third is an apiclient.errors.HttpError exception object if an HTTP
+            error occurred while processing the request, or None if no error
+            occurred.
```
nathanielmanistaatgoogle: Doesn't this doc string need a "Returns" section since the function returns a value?

eljobe: Done.  Sorry I screwed up the commit so badly in the process.

PR #109, opened 2015-06-23
Fix blogger API sample.

seppide:
Now the blogger API sample is actually working.


nathanielmanistaatgoogle:
Is it worth recording any more information about this in the commit log? When it broke, how it broke, why it broke, and/or whatnot?

Thanks for the fix!

seppide:
I have no idea why it was broken. It just didn't work..... ;) It looks like the product of some sweeping search-and-replace which went wrong and was never tested.

nathanielmanistaatgoogle:
Okay. Thanks much!

PR #114, opened 2015-07-16
Making http parameter in discovery.build_from_document optional.

jonparrott:
discovery.build()'s http parameter is optional, but if you call build_from_document directly it doesn't handle the case where http is None. Fixed that so now you can call ``discovery.build_from_document(doc, credentials=credentials)``.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -254,6 +254,9 @@ def build_from_document(
     A Resource object with methods for interacting with the service.
   """
 
+  if http is None:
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
How about test coverage? (This is my standard response to any contribution without test coverage, no matter how small a change.)

jonparrott:
Tests added for the case where you do provide http and where you don't provide http, covering all branches there.

nathanielmanistaatgoogle:
Thanks for the contribution!

PR #115, opened 2015-07-20
Stored Reason on HttpErrors

elibixby:
In the case where there are multiple reason strings for HttpErrors with the same code, end users may want to filter errors based on the reason string.


elibixby:
@jonparrott

elibixby:
Added a commit to change None handling. If reason is exposed, should use None instead of empty string.

elibixby:
Tests to green

jonparrott:
@elibixby Is this PR still applicable?

elibixby:
image

PR #117, opened 2015-07-23
Discovery: Treat empty nextPageToken field the same as it not existing

sondinhinfectious:
Sending nextPageToken out with an empty string to Google's DFA API will
raise a 500 error. This commit treats an empty nextPageToken field the
same as it not existing at all so that method.list_next() will not
attempt to retrieve a non-existent page of results.


nathanielmanistaatgoogle:

PR #119, opened 2015-07-30
return if the batch is empty

zeapo:
The batch execute when not given an http object will look into the given requests. If no request was given, it will raise a `ValueError("Missing a valid http object.")`.


zeapo:
I signed it!

nathanielmanistaatgoogle:
This looks alright, but what use case is this supporting? How did this situation arise for you?

zeapo:
We loop through a list of users and batch add them to a group. The users to add have to respect a certain criteria, and we expected that if none of them was to be added, to have nothing happen.

nathanielmanistaatgoogle:
Okay. Thanks for the contribution!

PR #125, opened 2015-08-13
Force HttpMock to read content from file as bytes

### LINKED ISSUE #124: HttpMock response content should be bytes but isn't (in Python 3)
Consider the following test code, to mock a 400 Bad Request error (as returned if the API key is invalid):

```python
http = HttpMock('bad_request.json', {'status': '400'})
response = some_api_call(http=http)
```

The expected behaviour here would be raising an ``HttpError``, however on Python 3 this results in a mysterious ``TypeError: HTTP content should be bytes`` bubbling up from the ``HttpError`` constructor.

My investigations led me to discover that:
* HTTP response content should always be bytes (confirmed [in a related pull request](https://github.com/google/google-api-python-client/pull/81#r27543602))
* ``HttpMock`` reads response content from files in text mode, so it gets a string rather than bytes on Python 3
* ``HttpError`` contains the only ``isinstance(..., bytes)`` check in the entire library, so testing with non-error status codes yields working tests
* 100% of test cases using ``HttpMock`` use status=200

As far as I can tell, this is an extremely simple fix: ``HttpMock`` should read the file in ``rb`` mode. (Simple) pull request with test and fix to follow.

alanbriolat:
Fixes #124.

I decided to write the test for the functionality I was actually trying to use, and which was previously not covered by tests: using ``HttpMock`` specifically to mock error responses. The side effect is that it exposes how ``HttpMock`` was not not reading content as bytes on Python 3 and causing a confusing exception to be thrown.

I think the fix speaks for itself...

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1463,7 +1463,7 @@ class HttpMock(object):
     if headers is None:
       headers = {'status': '200'}
     if filename:
-      f = open(filename, 'r')
+      f = open(filename, 'rb')
       self.data = f.read()
       f.close()
     else:
--- a/tests/data/bad_request.json
+++ b/tests/data/bad_request.json
@@ -1,0 +1,13 @@
+{
+ "error": {
+  "errors": [
+   {
+    "domain": "usageLimits",
+    "reason": "keyInvalid",
+    "message": "Bad Request"
+   }
+  ],
+  "code": 400,
+  "message": "Bad Request"
+ }
+}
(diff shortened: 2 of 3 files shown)


alanbriolat:
CLA has now been signed.

nathanielmanistaatgoogle:
A one-byte fix? Are we sure there's not something less drastic we could do instead? :-P

nathanielmanistaatgoogle:
Thanks for the contribution!

PR #131, opened 2015-09-04
Fixed a NameError on exception, added a unit test.

### LINKED ISSUE #130: NameError: name 'serviceName' is not defined
File "oauth2client/oauth2client/util.py", line 142, in positional_wrapper
  File "google-api-python-client/googleapiclient/discovery.py", line 193, in build
    content = _retrieve_discovery_doc(requested_url, http, cache_discovery, cache)
  File "googleapiclient/discovery.py", line 236, in _retrieve_discovery_doc
    raise UnknownApiNameOrVersion("name: %s  version: %s" % (serviceName,
NameError: name 'serviceName' is not defined

The same for `version` too, not passed to this function
Other http errors are also not handled.
I suggest you move out the codes inside function `_retrieve_discovery_doc`, and remove this function because it's only used once.
The code is messy

tmatsuo:
Fixes #130

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -29,6 +29,7 @@ __all__ = [
     ]
 
 from six import StringIO
+from six.moves import http_client
 from six.moves.urllib.parse import urlencode, urlparse, urljoin, \
   urlunparse, parse_qsl
 
@@ -190,7 +191,15 @@ def build(serviceName,
 
   requested_url = uritemplate.expand(discoveryServiceUrl, params)
 
-  content = _retrieve_discovery_doc(requested_url, http, cache_discovery, cache)
+  try:
+    content = _retrieve_discovery_doc(requested_url, http, cache_discovery,
+                                      cache)
+  except HttpError as e:
+    if e.resp.status == http_client.NOT_FOUND:
+      raise UnknownApiNameOrVersion("name: %s  version: %s" % (serviceName,
+                                                               version))
+    else:
+      raise e
 
   return build_from_document(content, base=discoveryServiceUrl, http=http,
       developerKey=developerKey, model=model, requestBuilder=requestBuilder,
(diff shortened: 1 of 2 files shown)


tmatsuo:
@nathanielmanistaatgoogle 

Sorry about not adding unit test for the exception clause.

googleapiclient/discovery.py
```diff
@@ -190,7 +190,15 @@ def build(serviceName,
 
   requested_url = uritemplate.expand(discoveryServiceUrl, params)
 
-  content = _retrieve_discovery_doc(requested_url, http, cache_discovery, cache)
+  try:
+    content = _retrieve_discovery_doc(requested_url, http, cache_discovery,
+                                      cache)
+  except HttpError as e:
+    if e.resp.status == 404:
```
nathanielmanistaatgoogle: httplib.NOT_FOUND rather than the literal value 404, please.

nathanielmanistaatgoogle:
Thanks for the fix; just the one nitpick comment.

googleapiclient/discovery.py
tmatsuo: Done, httplib is only available with python2, so I needed a hack for py3 though.

tmatsuo:
@nathanielmanistaatgoogle Thanks for the comment, PTAL

googleapiclient/discovery.py
nathanielmanistaatgoogle: Oof, no hack please. How about "from six.moves import http_client"? (https://pythonhosted.org/six/#module-six.moves)

tmatsuo: I see, done. Thanks

tmatsuo:
@nathanielmanistaatgoogle PTAL

nathanielmanistaatgoogle:
Thanks!

PR #132, opened 2015-09-08
Catch exception when path parameters not in message.

volnt:
Raise an exception when any of the path parameters is not present in the
input message of the endpoint.

Example :

```
@endpoints.method(message_types.VoidMessage, message_types.VoidMessage, path='/{parameter}')
def sample_endpoint(self, request):
    return VoidMessage()

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -57,6 +57,7 @@ from googleapiclient.errors import MediaUploadSizeError
 from googleapiclient.errors import UnacceptableMimeTypeError
 from googleapiclient.errors import UnknownApiNameOrVersion
 from googleapiclient.errors import UnknownFileType
+from googleapiclient.errors import ParameterRequiredError
 from googleapiclient.http import BatchHttpRequest
 from googleapiclient.http import HttpRequest
 from googleapiclient.http import MediaFileUpload
@@ -846,7 +847,10 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
     required = ''
     if arg in parameters.required_params:
       required = ' (required)'
-    paramdesc = methodDesc['parameters'][parameters.argmap[arg]]
+    try:
+      paramdesc = methodDesc['parameters'][parameters.argmap[arg]]
+    except KeyError:
+      raise ParameterRequiredError('Missing path parameter "%s" in input message' % arg)
     paramdoc = paramdesc.get('description', 'A parameter')
     if '$ref' in paramdesc:
       docs.append(
--- a/googleapiclient/errors.py
+++ b/googleapiclient/errors.py
@@ -107,6 +107,10 @@ class InvalidNotificationError(Error):
   """The channel Notification is invalid."""
   pass
 
+class ParameterRequiredError(Error):
+  """The url parameter is not present in the input message."""
+  pass
+
 class BatchError(HttpError):
   """Error occured during batch operations."""
 

### CODE REVIEW

**Comment by volnt** [comment]
I signed it!

**Comment by nathanielmanistaatgoogle** [comment]
@volnt: I'm not sure I understand this change - is "path parameters not in message" an illegal thing for applications to do, and this change simply improves the quality of the exception that is raised, or is it a legal thing to do, and the use of a custom error class is warranted because KeyError indicates a programming defect but `InvalidMessageError` indicates... something else? How about test coverage for this change - I think that that would help me see legitimate use cases in which the exception is raised.

@tmatsuo, do you want to look at this since it is discovery-related?

**Comment by volnt** [comment]
@nathanielmanistaatgoogle This change improves the quality of the exception that is raised. 

When you use the endpoint example I wrote above, the discovery module just throws "KeyError: u'parameter'". This exception doesn't indicate anything about why the exception is raised.

**Comment by volnt** [comment]
I will look into making a test to cover this change.

File: googleapiclient/discovery.py [hunk 1]
```diff
@@ -57,6 +57,7 @@
 from googleapiclient.errors import UnacceptableMimeTypeError
 from googleapiclient.errors import UnknownApiNameOrVersion
 from googleapiclient.errors import UnknownFileType
+from googleapiclient.errors import InvalidMessageError
```
tmatsuo: This should go just after InvalidJsonError

googleapiclient/errors.py
```diff
@@ -107,6 +107,10 @@ class InvalidNotificationError(Error):
   """The channel Notification is invalid."""
   pass
 
+class InvalidMessageError(Error):
```
tmatsuo: The name InvalidMessage is more generic than what the docstring says. Maybe you can change the docstring a bit to match the name?

tmatsuo:
@nathanielmanistaatgoogle It is in the discovery.py, but the change is actually in the generated method. I'll take a look anyways.

@volnt I think it's when you try to use your endpoint API from the client side. Maybe you can add a samplle code for such wrong usage?

googleapiclient/errors.py
tmatsuo: On the second thought, a name like ParameterRequiredError looks more appropriate. What do you think @volnt ?

volnt: Yes ParameterRequiredError looks great, I'll update the code with that.

volnt:
A sample code would be : 

```
@endpoints.method(message_types.VoidMessage, message_types.VoidMessage, path='/{parameter}', name='sample_endpoint')
def sample_endpoint(self, request):
    return VoidMessage()
```
```
>>> service = apiclient.discovery.build(sample_api, sample_version, discoveryServiceUrl=sample_discovery_url)
>>> service.sample().sample_endpoint(parameter="sample_parameter").execute()
 ...
 ...
KeyError: u'parameter'
```

With my change it outputs : 
```
 ...
 ...
TypeError: Missing path parameter "parameter" in input message
```

nathanielmanistaatgoogle:
Now that this has settled a bit: why isn't `ValueError` the appropriate exception to use? According to [its spec](https://docs.python.org/2/library/exceptions.html#exceptions.ValueError), `ValueError` indicates that a programming mistake has been made by passing particular values of arguments to a function. Isn't that the case here? If what we want is a more helpful message communicated to the user, can't we do that without having to define a new exception type?

tmatsuo:
@nathanielmanistaatgoogle Good catch. I agree.

@volnt I think your sample has `parameter` argument, which is not intended maybe?

JFYI this is a good talk: http://pyvideo.org/video/880/stop-writing-classes

pferate:
@tmatsuo Thanks for the video!  Jack did a similar talk at our Seattle Python meetup this summer, but it was more geared at names and namespaces, which he touched on here.

jonparrott:
@volnt I'm closing this PR due to inactivity. If you're still interested in getting this merged please let us know and we'll be happy to re-open it.

PR #133, opened 2015-09-10
implement exponential backoff and retry for BatchHttpRequest

jay0lee:


googleapiclient/http.py
```diff
@@ -1005,6 +1005,10 @@ def __init__(self, callback=None, batch_uri=None):
     # A map of id(Credentials) that have been refreshed.
     self._refreshed_credentials = {}
 
+    # Stubs for testing.
```
nathanielmanistaatgoogle: This commit doesn't seem to include test coverage?

nathanielmanistaatgoogle:
For my own understanding of the scope of this change: how severe is the problem of the library not having this functionality and why is this the right change to make to address the problem you're having?

lickdragon:
Hi nathanielmanistaatgoogle,
   HttpError 500 certainly affect me when using duplicity backing up to google drive.
   StackOverflow also has a question on it.  One answer apparently by a google employee http://stackoverflow.com/a/12640475 made it seem like HttpError 500's are an expected part of interacting with Google Drive (web services in general).  So I'd expect implementing a retry for HttpError 500 to be useful to many people.
   I'd like to test the patch, but am a little clueless.  How do I download and install the patched google-drive-api-client ?  (I've only used pip ...)

pferate:
@lickdragon, you can download the code and install it with the `setup.py` script.

`python setup.py install`

Or use `git+https://github.com/jay0lee/google-api-python-client.git` to install the version from @jay0lee's fork.

`pip install --upgrade git+https://github.com/jay0lee/google-api-python-client.git`

Although it looks like @jay0lee's fork is 129 commits behind the current version, so I'm not sure all of the differences (I know lack of Python3 for sure)

lickdragon:
Tnanks Pat!
I'll test it out when I get a chance:  Probably best to use the official 
repository then apply the jay0lee patches.

C.

rhan-mentad:
Why don't you use funcy's retry decorator instead of implementing it youself?

jonparrott:
@jay0lee I'm closing this PR as inactive. If you're still interested in getting this merged, please open a new PR.

Issue #126, opened 2015-08-19
"Issue tracker" link is broken on https://google.github.io/google-api-python-client/contributing.html


harisibrahimkv:
The "issue tracker" link points to https://github.com/google/google/google-api-python-client/issues

It should point to https://github.com/google/google-api-python-client/issues

nathanielmanistaatgoogle:
Thanks for the report! Now that I spend a while looking at it... something's wrong with google-api-python-client's documentation process; I've elaborated a bit in #75. In the meantime perhaps contribute a pull request against our `gh-pages` branch?

pferate:
Created PR #143 to resolve this.

bantini:
Closing because this issue has been resolved.

PR #143, opened 2015-10-12
Removing redundant name in GH paths

pferate:
PR for resolving issue #126

### DIFF
--- a/contributing.html
+++ b/contributing.html
@@ -18,7 +18,7 @@
         <h1>Contributing to google/google-api-python-client</h1>
         <p></p>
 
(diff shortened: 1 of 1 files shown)


nathanielmanistaatgoogle:
Of course the change looks fine content-wise and should be merged, but I'm disturbed that the tests didn't run - any idea how that happened? I'm not aware of us having smarts that say "don't run if it's only a documentation change", but maybe we do?

pferate:
Travis ignores the gh-pages branch unless it is explicitly called out in
the Travis config file.

nathanielmanistaatgoogle:
Right. Monday morning brain. :-(

PR #146, opened 2015-10-13
Use the maintained uritemplate.py

graingert:


graingert:
`I signed it!`

nathanielmanistaatgoogle:
This smells like a change that is at least as political as it is technical. What's the story? Why was uritemplate forked? When's it going to be unforked?

pferate:
I agree @nathanielmanistaatgoogle.  It doesn't look like it's a fork, just a separate implementation of the same RFC.  On PyPI, they were both last updated 2 years ago, although they both have updated on GitHub since then.  I would recommend staying with the code base maintained by the RFC authors, unless there is a compelling reason not to.

graingert:
@pferate it looks like they're both alive again let's stick with the current package

PR #147, opened 2015-10-13
Fix non-resumable binary uploads on Python 3

### LINKED ISSUE #145: Failure when uploading binary files on Python 3 when resumable=False
Seems like `MediaFileUpload` (or something else deep inside) is failing to handling binary files properly on Python 3.  My guess is that the contents of the file is mistakenly passed as a Unicode string (rather than as a byte-string) into `http.client`, which dies as soon as it encounters a non-Latin-1 character.  This only happens when `resumable=False`.

Code to reproduce the problem:

~~~py
import apiclient
service = apiclient.discovery.build("drive", "v2", …)
filename = "some-file.tar.xz"
media = apiclient.http.MediaFileUpload(filename)
service.files().insert(body={"title": os.path.basename(filename)}, media_body=media).execute()
~~~

Typical error:

~~~
  …
  File "/usr/lib/python3.5/site-packages/oauth2client/util.py", line 142, in positional_wrapper
  File "/usr/lib/python3.5/site-packages/googleapiclient/http.py", line 722, in execute
    body=self.body, headers=self.headers)
  File "/usr/lib/python3.5/site-packages/oauth2client/client.py", line 589, in new_request
    redirections, connection_type)
  File "/usr/lib/python3.5/site-packages/httplib2/__init__.py", line 1313, in request
    (response, content) = self._request(conn, authority, uri, request_uri, method, body, headers, redirections, cachekey)
  File "/usr/lib/python3.5/site-packages/httplib2/__init__.py", line 1063, in _request
    (response, content) = self._conn_request(conn, request_uri, method, body, headers)
  File "/usr/lib/python3.5/site-packages/httplib2/__init__.py", line 987, in _conn_request
    conn.request(method, request_uri, body, headers)
  File "/usr/lib/python3.5/http/client.py", line 1083, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.5/http/client.py", line 1127, in _send_request
    body = body.encode('iso-8859-1')
UnicodeEncodeError: 'latin-1' codec can't encode character '\ufffd' in position 350: ordinal not in range(256)
~~~

(Using Python 3.5.0 on Arch Linux 4.2.2.)

Rufflewind:
1. `Generator` and `StringIO` are replaced by `BytesGenerator` and `BytesIO`.  If `BytesGenerator` doesn't exist (as is the case in Python 2), fall back to Generator.

 2. `BytesGenerator` is buggy (see [#18886](https://bugs.python.org/issue18886) and [#19003](https://bugs.python.org/issue19003)) and corrupts `\r` into `\n`.  To work around this, we implement a patched version of `BytesGenerator` that replaces `._write_lines` with just `.write`.

Note: `BytesGenerator` was introduced in Python 3.2.  This is OK since the library already demands 3.3+.

The `test_multipart_media_good_upload` has been updated to reflect the change.  It is also stricter now, as it matches the entire request body against the expected form.

Fixes #145.

---

This ensures that `body` is of type `bytes` rather than `str`.

I did some tests to make sure the new code is still sensible: the flattened MIME of a simple upload of a random 1MiB file on Python 3.5.0 after the fix is identical to that on Python 2.7.10 before and after the fix (modulo `dict` rearrangements and MIME boundaries).

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -28,14 +28,17 @@ __all__ = [
     'key2param',
     ]
 
-from six import StringIO
+from six import BytesIO
 from six.moves import http_client
 from six.moves.urllib.parse import urlencode, urlparse, urljoin, \
   urlunparse, parse_qsl
 
 # Standard library imports
 import copy
-from email.generator import Generator
+try:
+  from email.generator import BytesGenerator
+except ImportError:
+  from email.generator import Generator as BytesGenerator
 from email.mime.multipart import MIMEMultipart
 from email.mime.nonmultipart import MIMENonMultipart
 import json
@@ -102,6 +105,10 @@ STACK_QUERY_PARAMETER_DEFAULT_VALUE = {'type': 'string', 'location': 'query'}
 # Library-specific reserved words beyond Python keywords.
 RESERVED_WORDS = frozenset(['body'])
 
+# patch _write_lines to avoid munging '\r' into '\n'
+# ( https://bugs.python.org/issue18886 https://bugs.python.org/issue19003 )
+class _BytesGenerator(BytesGenerator):
+  _write_lines = BytesGenerator.write
 
 def fix_method_name(name):
   """Fix method names to avoid reserved word conflicts.
@@ -797,8 +804,8 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
           msgRoot.attach(msg)
           # encode the body: note that we can't use `as_string`, because
           # it plays games with `From ` lines.
-          fp = StringIO()
-          g = Generator(fp, mangle_from_=False)
+          fp = BytesIO()
+          g = _BytesGenerator(fp, mangle_from_=False)
           g.flatten(msgRoot, unixfrom=False)
           body = fp.getvalue()
 
--- a/tests/test_discovery.py
+++ b/tests/test_discovery.py
@@ -35,6 +35,7 @@ import itertools
 import json
 import os
 import pickle
+import re
 import sys
 import unittest2 as unittest
 
@@ -787,7 +788,21 @@ class Discovery(unittest.TestCase):
     request = zoo.animals().insert(media_body=datafile('small.png'), body={})
     self.assertTrue(request.headers['content-type'].startswith(
         'multipart/related'))
-    self.assertEquals('--==', request.body[0:4])
+    with open(datafile('small.png'), 'rb') as f:
+      contents = f.read()
+    boundary = re.match(b'--=+([^=]+)', request.body).group(1)
+    self.assertEqual(
+      request.body.rstrip(b"\n"), # Python 2.6 does not add a trailing \n
+      b'--===============' + boundary + b'==\n' +
+      b'Content-Type: application/json\n' +
+      b'MIME-Version: 1.0\n\n' +
+      b'{"data": {}}\n' +
+      b'--===============' + boundary + b'==\n' +
+      b'Content-Type: image/png\n' +
+      b'MIME-Version: 1.0\n' +
+      b'Content-Transfer-Encoding: binary\n\n' +
+      contents +
+      b'\n--===============' + boundary + b'==--')
     assertUrisEqual(self,
         'https://www.googleapis.com/upload/zoo/v1/animals?uploadType=multipart&alt=json',
         request.uri)


Rufflewind:
I signed it!

nathanielmanistaatgoogle:
What hope is there for test coverage for this change? And what's up with the test failures seen for the current draft of your change?

Rufflewind:
The broken test is now fixed.  It is also stricter now and should serve as a regression test for this bug.

nathanielmanistaatgoogle:
This looks correct to me.

@tmatsuo would you mind being a second pair of eyes on this? The try/except import and monkeypatching are giving me the heebie-jeebies.

googleapiclient/discovery.py
```diff
@@ -797,9 +800,17 @@ def method(self, **kwargs):
...
-          g = Generator(fp, mangle_from_=False)
+          fp = BytesIO()
+          g = BytesGenerator(fp, mangle_from_=False)
+          # monkeypatch _write_lines to avoid munging '\r' into '\n'
+          # ( https://bugs.python.org/issue18886 );
+          # since Python 2 does not have _write_lines so this is a no-op
+          write_lines = getattr(BytesGenerator, "_write_lines", None)
+          if write_lines:
+            BytesGenerator._write_lines = BytesGenerator.write
           g.flatten(msgRoot, unixfrom=False)
```
tmatsuo: What happens if g.flatten throws an Exception? I assume BytesGenerator.write will still be patched.

Rufflewind:
> What happens if g.flatten throws an Exception? I assume BytesGenerator.write will still be patched.

Fixed.

~~~diff
-          g.flatten(msgRoot, unixfrom=False)
-          if write_lines:
-            BytesGenerator._write_lines = write_lines
+          try:
+            g.flatten(msgRoot, unixfrom=False)
+          finally:
+            if write_lines:
+              BytesGenerator._write_lines = write_lines
~~~

tmatsuo:
I'm still worrying when the code is used in multi threaded code. I think some threads might pick the patched version as `original`.

nathanielmanistaatgoogle:
Let me ask an alarmingly naïve question that reveals how I see the world with precious, innocent, childlike wonder: how has [Python bug 18886](https://bugs.python.org/issue18886) stayed open so long if it is so easy to work around?

Rufflewind:
> I'm still worrying when the code is used in multi threaded code. I think some threads might pick the patched version as `original`.

Found a better way to do this that sidesteps the issue :)

> Let me ask an alarmingly naïve question that reveals how I see the world with precious, innocent, childlike wonder: how has [Python bug 18886](https://bugs.python.org/issue18886) stayed open so long if it is so easy to work around?

There may be a subtlety: some applications may have a preference for `\r\n` vs `\n` in the header lines, which may add slight complications for the fix on their end.  This is does not matter for googleapiclient since the Google API appears to tolerate `\n` just fine.

The same bug has also been reported as [#19003](https://bugs.python.org/issue19003), which includes a patch.

googleapiclient/discovery.py
```diff
@@ -102,6 +105,10 @@
 # Library-specific reserved words beyond Python keywords.
 RESERVED_WORDS = frozenset(['body'])
 
+# patch _write_lines to avoid munging '\r' into '\n'
+# ( https://bugs.python.org/issue18886 )
```
nathanielmanistaatgoogle: If you feel that 19003 is a duplicate of 18886, put both on this line?

Rufflewind:
> nathanielmanistaatgoogle: If you feel that 19003 is a duplicate of 18886, put both on this line?

Done.

tmatsuo:
Thanks, LGTM

nathanielmanistaatgoogle:
Thank you for this conversation and for your contribution to the library!

chromafunk:
Is the available master the one that installs via pip ? Cause I am having exactly the original issue that lead to this patch !

nathanielmanistaatgoogle:
@chromafunk: unfortunately no, we're overdue for a release. Perhaps in the next week? Fingers crossed?

chromafunk:
@nathanielmanistaatgoogle I have modified discovery.py with the patch by @Rufflewind myself but the problem persist when using resumable = False, do you want me to help on this ? I can dump output if you want it.

naoyak:
A new release would be terrific!

PR #149, opened 2015-10-19
Update README.md

jrogers1126:
make groups in groupsettings plural (groupssetting)


nathanielmanistaatgoogle:
I like the content of this change but can't integrate it into the codebase without your having signed the CLA.

jrogers1126:
I've signed the CLA already.

Issue #158, opened 2015-11-25
Please retry on network errors


jurekPapiorek:
The method HttpRequets.execute has an option to retry on 500s. Please also provide an option to retry on network errors (we've seen the method throw each of {socket.timeout, ssl.SSLError}). Also please provide an option to auto-retry on rate limit exceeded errors.

Issue #97, opened 2015-04-29
SSLError: EOF occurred in violation of protocol


harishpendela:
I am using google-api-python-client_v1.2 on python 2.7.3. Often I see the following error while trying to build a drive service object. This is the stack trace:

drive_service = build('drive', 'v2', http=http)
File "/local/lib/python2.7/site-packages/google_api_python_client-1.2-py2.7.egg/oauth2client/util.py", line 132, in positional_wrapper
   return wrapped(_args, *_kwargs)
 File "/local/lib/python2.7/site-packages/google_api_python_client-1.2-py2.7.egg/apiclient/discovery.py", line 192, in build
resp, content = http.request(requested_url)
 File "/local/lib/python2.7/site-packages/google_api_python_client-1.2-py2.7.egg/oauth2client/util.py", line 132, in positional_wrapper
return wrapped(_args, *_kwargs)
 File "/local/lib/python2.7/site-packages/google_api_python_client-1.2-py2.7.egg/oauth2client/client.py", line 475, in new_request
self._refresh(request_orig)
 File "/local/lib/python2.7/site-packages/google_api_python_client-1.2-py2.7.egg/oauth2client/client.py", line 653, in _refresh
   self._do_refresh_request(http_request)
 File "/local/lib/python2.7/site-packages/google_api_python_client-1.2-py2.7.egg/oauth2client/client.py", line 682, in _do_refresh_request
   self.token_uri, method='POST', body=body, headers=headers)
File "/local/lib/python2.7/site-packages/httplib2/__init__.py", line 1593, in request
(response, content) = self._request(conn, authority, uri, request_uri, method, body, headers, redirections, cachekey)
File "/local/lib/python2.7/site-packages/httplib2/__init__.py", line 1335, in _request
(response, content) = self._conn_request(conn, request_uri, method, body, headers)
 File "/local/lib/python2.7/site-packages/httplib2/__init__.py", line 1257, in _conn_request
   self.disable_ssl_certificate_validation, self.ca_certs)
File "/local/lib/python2.7/site-packages/httplib2/__init__.py", line 80, in _ssl_wrap_socket
cert_reqs=cert_reqs, ca_certs=ca_certs)
 File "/usr/lib/python2.7/ssl.py", line 381, in wrap_socket
   ciphers=ciphers)
File "/usr/lib/python2.7/ssl.py", line 143, in __init__
   self._sslobj.do_handshake()
[...]

PR #159, opened 2015-12-15
Handle SSL errors with retries

sergiyby:
This was already reviewed once at https://codereview.chromium.org/1522953002.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -36,6 +36,7 @@ import logging
 import mimetypes
 import os
 import random
+import ssl
 import sys
 import time
 import uuid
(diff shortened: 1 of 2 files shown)


sergiyby:
This is going to address issues #71, #97, #158.

googleapiclient/http.py
```diff
@@ -546,14 +547,21 @@ def next_chunk(self, num_retries=0):
         }
     http = self._request.http
 
+    resp = None
```
nathanielmanistaatgoogle: It's not clear to me why this line is necessary? What breaks if it is removed?

googleapiclient/http.py
```diff
@@ -799,16 +815,24 @@ def next_chunk(self, http=None, num_retries=0):
...
-                                     headers=start_headers)
+              'Retry #%d for resumable URI request: %s %s%s'
+              % (retry_num, self.method, self.uri,
+                 ', following status: %d' % resp.status if resp else ''))
+
+        try:
+          resp, content = http.request(self.uri, method=self.method,
+                                       body=self.body,
+                                       headers=start_headers)
+        except ssl.SSLError:
```
nathanielmanistaatgoogle: This logic appears nearly identically in three places - please factor it out into a module-scope module-private helper function?

nathanielmanistaatgoogle:

googleapiclient/http.py
sergiyby: When SSLError happens, resp is not assigned and thus not defined which breaks line 556 which tries to use it in a condition.

googleapiclient/http.py
sergiyby: Done.

sergiyby:
I've been using a mixture of accounts to create CLs and this pull request, which seems to have confused the CLA check. I'll re-create the pull request from the same account as the author.

PR #169, opened 2016-02-05

ethanbao:
Fallback to use new discovery uri pattern when the old one fails. Some Google API are now using the new Discovery url style.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -82,6 +82,8 @@ URITEMPLATE = re.compile('{[^}]*}')
 VARNAME = re.compile('[a-zA-Z0-9_-]+')
 DISCOVERY_URI = ('https://www.googleapis.com/discovery/v1/apis/'
                  '{api}/{apiVersion}/rest')
+DISCOVERY_URI2 = ('https://{api}}.googleapis.com/$discovery/rest?'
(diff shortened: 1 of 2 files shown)


googleapiclient/discovery.py
```diff
@@ -82,6 +82,8 @@
 VARNAME = re.compile('[a-zA-Z0-9_-]+')
 DISCOVERY_URI = ('https://www.googleapis.com/discovery/v1/apis/'
                  '{api}/{apiVersion}/rest')
+DISCOVERY_URI2 = ('https://{api}}.googleapis.com/$discovery/rest?'
```
nathanielmanistaatgoogle: Yuck no. Please name this field `V2_DISCOVERY_URI` and introduce a `V1_DISCOVERY_URI` to match it.

Unfortunately we have to maintain `DISCOVERY_URI` indefinitely as a public field of this module, but at least it can be an alias for `V1_DISCOVERY_URI`.

Actually, it can be an alias of `V2_DISCOVERY_URI`, can't it? Is that the URI to which you want most users of this library to go?

nathanielmanistaatgoogle:
Please address the CLA complaint by both authoring and committing from either your google.com account or an account that has signed Google's CLA.

PR #173, opened 2016-02-08

ethanbao:


googleapiclient/discovery.py
```diff
@@ -153,7 +157,7 @@ def key2param(key):
 def build(serviceName,
           version,
           http=None,
-          discoveryServiceUrl=DISCOVERY_URI,
+          discoveryServiceUrl=V1_DISCOVERY_URI,
```
nathanielmanistaatgoogle: Why shouldn't this value be changed to `V2_DISCOVERY_URI`?

googleapiclient/discovery.py
```diff
@@ -196,21 +200,23 @@ def build(serviceName,
...
+    except HttpError as e:
+      if e.resp.status == http_client.NOT_FOUND:
+        continue
+      else:
+        raise e
 
-  return build_from_document(content, base=discoveryServiceUrl, http=http,
-      developerKey=developerKey, model=model, requestBuilder=requestBuilder,
-      credentials=credentials)
+  raise UnknownApiNameOrVersion("name: %s  version: %s" % (serviceName,
```
nathanielmanistaatgoogle: Line-wrap after the first opening parenthesis:
```
raise UnknownApiNameOrVersion(
    "name: %s  version: %s" % (serviceName, version))
```
.

googleapiclient/discovery.py
```diff
@@ -196,21 +200,23 @@ def build(serviceName,
...
-  try:
-    content = _retrieve_discovery_doc(requested_url, http, cache_discovery,
-                                      cache)
-  except HttpError as e:
-    if e.resp.status == http_client.NOT_FOUND:
-      raise UnknownApiNameOrVersion("name: %s  version: %s" % (serviceName,
-                                                               version))
-    else:
-      raise e
+  for discovery_url in [discoveryServiceUrl, V2_DISCOVERY_URI]:
```
nathanielmanistaatgoogle: The shape of this code may change in response to my question about the default value of the `discoveryServiceUrl` parameter, but with respect to the current draft: why try the old thing and then fall back to the new thing? Why not try the new thing first and then fall back to the old thing if necessary?

googleapiclient/discovery.py
```diff
@@ -82,6 +82,10 @@
 VARNAME = re.compile('[a-zA-Z0-9_-]+')
 DISCOVERY_URI = ('https://www.googleapis.com/discovery/v1/apis/'
                  '{api}/{apiVersion}/rest')
+V1_DISCOVERY_URI = DISCOVERY_URI
+V2_DISCOVERY_URI = ('https://{api}.googleapis.com/$discovery/rest?'
```
nathanielmanistaatgoogle: If you line-wrap immediately after the opening parenthesis can you fit your entire string on the following line without breaking it?

Implicit string concatenation is one of Python's most dangerous problems and one should always try to avoid it.

googleapiclient/discovery.py
nathanielmanistaatgoogle: Never construct a list when a tuple will do. Can this be
```
for discovery_url in (discoveryServiceUrl, V2_DISCOVERY_URI,):
```
?

nathanielmanistaatgoogle:
Please squash commits with each round of code review.

Also please ensure that your commit log message conforms to [these guidelines](http://chris.beams.io/posts/git-commit/#seven-rules).

PR #174, opened 2016-02-08
__name__ loggers

emmett9001:
This pull request moves logging into loggers that are specific to the current `__name__`. This allows code that uses the API client to set the logging level for the client without affecting other logging happening on the root logger.


googleapiclient/discovery_cache/__init__.py
```diff
@@ -19,6 +19,9 @@
 import logging
 import datetime
 
+
+logger = logging.getLogger(__name__)
```
nathanielmanistaatgoogle: Since this is a module-scope constant its name should be styled as `LOGGER`, right? (Here and throughout this change.)

nathanielmanistaatgoogle:
Please squash commits with each round of code review.

googleapiclient/discovery_cache/__init__.py
emmett9001: I was remaining consistent with the style [here](https://github.com/google/google-api-python-client/blob/30125120b48860d00ba4e96d56a86d9d02a90960/googleapiclient/discovery_cache/file_cache.py#L37), but I can make this change if this repo prefers the all-caps style.

nathanielmanistaatgoogle:
@jonparrott, what do you think of this?

jonparrott:
This is fine, but we might need to attach a nullhandler to the `oauth2client` logger ala urllib3: 

https://github.com/shazow/urllib3/blob/master/urllib3/__init__.py#L23

nathanielmanistaatgoogle:
What dark sorcery is that?

jonparrott:
It's not sorcery, it's just the right thing to do:

https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library

nathanielmanistaatgoogle:
Huh. Well that seems... curious.

Should @emmett9001 amend his commit to satisfy that guidance? (I suspect yes; @emmett9001, what do you think?)

jonparrott:
Yes, please.

emmett9001:
Just to clarify - are you suggesting that [this code](https://github.com/shazow/urllib3/blob/f0d98555ce6e3eeb379a8a659dff9d7f9ff42348/urllib3/__init__.py#L24-L30) should be added somewhere in my existing diff? It's not clear where this should go.

jonparrott:
Yes, it should go in `googleapiclient/__init__.py`.

emmett9001:
See my updated diff for the `NullHandler` logic.

jonparrott:
Looks good, except you this line: https://github.com/shazow/urllib3/blob/f0d98555ce6e3eeb379a8a659dff9d7f9ff42348/urllib3/__init__.py#L53

Once that's in and you've squashed commits, this LGTM.

emmett9001:
I messed up the history by rebasing and merging master, so I've reopened a clean version of this pull request as #206. Please comment further there and close this one if you'd like.

nathanielmanistaatgoogle:
In the future please keep to one pull request for a desired change to the codebase; it makes for better forensics to have all the code review recorded in one place.

emmett9001:
Will do. My previous pull request was corrupted by a bad merge - I'll be
sure to take care next time.

Issue #162, opened 2015-12-25
MediaIoBaseDownload class is not able to handle "Too Many Requests" error


nikesh-mahalka:
I am using ubuntu 14.04.03 server and installed google library with below command:
sudo pip install --upgrade google-api-python-client

google-api-python-client library installed by above command is some commits behind the master branch.

I changed if condition for handling this error in line number 553 in /usr/local/lib/python2.7/dist-packages/googleapiclient/http.py
`if resp.status < 500 and resp.status not in [err1, err2]:` 

And i am able to download a google object by retry mechanism on getting error 429 "Too Many Requests" error.

nikesh-mahalka:
```
from apiclient import http
from apiclient import discovery
from oauth2client import client
import six
credentials = client.GoogleCredentials.from_stream("path-to-json-file")
service = discovery.build('storage', 'v1', credentials=credentials)
req = service.objects().get_media(
            bucket="gcsbucket",
            object='File.exe',)
fh = six.BytesIO()
downloader = http.MediaIoBaseDownload(
            fh, req, chunksize=32768)
done = False
chunk_number = 0
while not done:
    status, done = downloader.next_chunk(num_retries=5)
    chunk_number = chunk_number + 1
    print done, chunk_number
```

Above code was giving too many request error 429, which is now not coming after modifying
https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L99 to
if resp.status < 500 and resp.status not in [err1, err2]:

nikesh-mahalka:
same for uploading a object 
https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L707

retry will not work in "Too many requests error 429".

mcdonc:
This code has been through so many changes by now that the original line of code mentioned by OP no longer exists, and I'm not keen on doing the archaelogy to find the version that was in use when the bug was submitted to do a 3-way comparison.  I'm going to close this as a result; if you find the problem continues to exist in modern versions of google-api-python-client, please reopen.

PR #177, opened 2016-02-12
Add retry on rate limiting API responses and timeouts. Fixes #170 and #162.

### LINKED ISSUE #170: Retry on HTTP 403 with "Rate Limit Exceeded"
The library should retry on such API responses when num_retries is set, instead of failing.

Currently, it fails:
`HttpError: <HttpError 403 when requesting https://www.googleapis.com/calendar/v3/calendars/user%40domain.com/events/import?supportsAttachments=true&alt=json returned "Rate Limit Exceeded">`

eesheesh:
Retry on rate-limiting API responses, based on https://developers.google.com/google-apps/calendar/v3/errors and https://tools.ietf.org/html/rfc6585#section-4.

Also retry on network timeout errors.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -20,6 +20,7 @@ actuall HTTP request.
 """
 from __future__ import absolute_import
 import six
+from six.moves import http_client
 from six.moves import range
 
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
@@ -36,6 +37,7 @@ import logging
 import mimetypes
 import os
 import random
+import socket
 import ssl
 import sys
 import time
@@ -63,6 +65,51 @@ DEFAULT_CHUNK_SIZE = 512*1024
 
 MAX_URI_LENGTH = 2048
 
+_TOO_MANY_REQUESTS = 429
+
+
+def _should_retry_response(resp_status, content):
+  """Determines whether a response should be retried.
+
+  Args:
+    resp_status: The response status received.
+    content: The response content body. 
+
+  Returns:
+    True if the response should be retried, otherwise False.
+  """
+  # Retry on 5xx errors.
+  if resp_status >= 500:
+    return True
+
+  # Retry on 429 errors.
+  if resp_status == _TOO_MANY_REQUESTS:
+    return True
+
+  # For 403 errors, we have to check for the `reason` in the response to
+  # determine if we should retry.
+  if resp_status == six.moves.http_client.FORBIDDEN:
+    # If there's no details about the 403 type, don't retry.
+    if not content:
+      return False
+
+    # Content is in JSON format.
+    try:
+      data = json.loads(content.decode('utf-8'))
+      reason = data['error']['errors'][0]['reason']
+    except (UnicodeDecodeError, ValueError, KeyError):
+      LOGGER.warning('Invalid JSON content from response: %s', content)
+      return False
+
+    LOGGER.warning('Encountered 403 Forbidden with reason "%s"', reason)
+
+    # Only retry on rate limit related failures.
+    if reason in ('userRateLimitExceeded', 'rateLimitExceeded', ):
+      return True
+
+  # Everything else is a success or non-retriable so break.
+  return False
+
 
 def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
                    **kwargs):
@@ -84,21 +131,37 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
     resp, content - Response from the http request (may be HTTP 5xx).
   """
   resp = None
+  content = None
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
-      sleep(rand() * 2**retry_num)
+      # Sleep before retrying.
+      sleep_time = rand() * 2 ** retry_num
       LOGGER.warning(
-          'Retry #%d for %s: %s %s%s' % (retry_num, req_type, method, uri,
-          ', following status: %d' % resp.status if resp else ''))
+          'Sleeping %.2f seconds before retry %d of %d for %s: %s %s, after %s',
+          sleep_time, retry_num, num_retries, req_type, method, uri,
+          resp.status if resp else exception)
+      sleep(sleep_time)
 
     try:
+      exception = None
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
-      if retry_num == num_retries:
+    # Retry on SSL errors and socket timeout errors.
+    except ssl.SSLError as ssl_error:
+      exception = ssl_error
+    except socket.error as socket_error:
+      # errno's contents differ by platform, so we have to match by name.
+      if socket.errno.errorcode.get(socket_error.errno) not in (
+          'WSAETIMEDOUT', 'ETIMEDOUT', ):
         raise
+      exception = socket_error
+
+    if exception:
+      if retry_num == num_retries:
+        raise exception
       else:
         continue
-    if resp.status < 500:
+
+    if not _should_retry_response(resp.status, content):
       break
 
   return resp, content
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -87,7 +87,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
       sleep(rand() * 2**retry_num)
-      LOGGER.warning(
+      logging.warning(
           'Retry #%d for %s: %s %s%s' % (retry_num, req_type, method, uri,
           ', following status: %d' % resp.status if resp else ''))
 
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -87,7 +87,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
       sleep(rand() * 2**retry_num)
-      logging.warning(
+      LOGGER.warning(
           'Retry #%d for %s: %s %s%s' % (retry_num, req_type, method, uri,
           ', following status: %d' % resp.status if resp else ''))
 
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -31,9 +31,11 @@ from six.moves.urllib.parse import urlencode
 # Do not remove the httplib2 import
 import httplib2
 import logging
+import mock
 import os
 import unittest2 as unittest
 import random
+import socket
 import ssl
 import time
 
@@ -102,7 +104,7 @@ class MockCredentials(Credentials):
     headers['authorization'] = self._bearer_token + ' ' + str(self._refreshed)
 
 
-class HttpMockWithSSLErrors(object):
+class HttpMockWithErrors(object):
   def __init__(self, num_errors, success_json, success_data):
     self.num_errors = num_errors
     self.success_json = success_json
@@ -113,7 +115,45 @@ class HttpMockWithSSLErrors(object):
       return httplib2.Response(self.success_json), self.success_data
     else:
       self.num_errors -= 1
-      raise ssl.SSLError()
+      if self.num_errors == 1:
+        raise ssl.SSLError()
+      else:
+        if PY3:
+          ex = TimeoutError()
+        else:
+          ex = socket.error()
+        # Initialize the timeout error code to the platform's error code.
+        try:
+          # For Windows:
+          ex.errno = socket.errno.WSAETIMEDOUT
+        except AttributeError:
+          # For Linux/Mac:
+          ex.errno = socket.errno.ETIMEDOUT
+        # Now raise the correct timeout error.
+        raise ex
+
+
+class HttpMockWithNonRetriableErrors(object):
+  def __init__(self, num_errors, success_json, success_data):
+    self.num_errors = num_errors
+    self.success_json = success_json
+    self.success_data = success_data
+
+  def request(self, *args, **kwargs):
+    if not self.num_errors:
+      return httplib2.Response(self.success_json), self.success_data
+    else:
+      self.num_errors -= 1
+      ex = socket.error()
+      # Initialize the timeout error code to the platform's error code.
+      try:
+        # For Windows:
+        ex.errno = socket.errno.WSAECONNREFUSED
+      except AttributeError:
+        # For Linux/Mac:
+        ex.errno = socket.errno.ECONNREFUSED
+      # Now raise the correct timeout error.
+      raise ex
 
 
 DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
@@ -409,8 +449,8 @@ class TestMediaIoBaseDownload(unittest.TestCase):
 
     self.assertEqual(self.fd.getvalue(), b'123')
 
-  def test_media_io_base_download_retries_ssl_errors(self):
-    self.request.http = HttpMockWithSSLErrors(
+  def test_media_io_base_download_retries_connection_errors(self):
+    self.request.http = HttpMockWithErrors(
         3, {'status': '200', 'content-range': '0-2/3'}, b'123')
 
     download = MediaIoBaseDownload(
@@ -593,6 +633,51 @@ Content-Length: 14
 ETag: "etag/pony"\r\n\r\n{"foo": 42}
 --batch_foobarbaz--"""
 
+
+USER_RATE_LIMIT_EXCEEDED_RESPONSE = """{
+ "error": {
+  "errors": [
+   {
+    "domain": "usageLimits",
+    "reason": "userRateLimitExceeded",
+    "message": "User Rate Limit Exceeded"
+   }
+  ],
+  "code": 403,
+  "message": "User Rate Limit Exceeded"
+ }
+}"""
+
+
+RATE_LIMIT_EXCEEDED_RESPONSE = """{
+ "error": {
+  "errors": [
+   {
+    "domain": "usageLimits",
+    "reason": "rateLimitExceeded",
+    "message": "Rate Limit Exceeded"
+   }
+  ],
+  "code": 403,
+  "message": "Rate Limit Exceeded"
+ }
+}"""
+
+
+NOT_CONFIGURED_RESPONSE = """{
+ "error": {
+  "errors": [
+   {
+    "domain": "usageLimits",
+    "reason": "accessNotConfigured",
+    "message": "Access Not Configured"
+   }
+  ],
+  "code": 403,
+  "message": "Access Not Configured"
+ }
+}"""
+
 class Callbacks(object):
   def __init__(self):
     self.responses = {}
@@ -622,10 +707,22 @@ class TestHttpRequest(unittest.TestCase):
     self.assertEqual(method, http.method)
     self.assertEqual(str, type(http.method))
 
-  def test_retry_ssl_errors_non_resumable(self):
+  def test_no_retry_connection_errors(self):
     model = JsonModel()
     request = HttpRequest(
-        HttpMockWithSSLErrors(3, {'status': '200'}, '{"foo": "bar"}'),
+        HttpMockWithNonRetriableErrors(1, {'status': '200'}, '{"foo": "bar"}'),
+        model.response,
+        u'https://www.example.com/json_api_endpoint')
+    request._sleep = lambda _x: 0  # do nothing
+    request._rand = lambda: 10
+    with self.assertRaises(socket.error):
+      response = request.execute(num_retries=3)
+
+
+  def test_retry_connection_errors_non_resumable(self):
+    model = JsonModel()
+    request = HttpRequest(
+        HttpMockWithErrors(3, {'status': '200'}, '{"foo": "bar"}'),
         model.response,
         u'https://www.example.com/json_api_endpoint')
     request._sleep = lambda _x: 0  # do nothing
@@ -633,7 +730,7 @@ class TestHttpRequest(unittest.TestCase):
     response = request.execute(num_retries=3)
     self.assertEqual({u'foo': u'bar'}, response)
 
-  def test_retry_ssl_errors_resumable(self):
+  def test_retry_connection_errors_resumable(self):
     with open(datafile('small.png'), 'rb') as small_png_file:
       small_png_fd = BytesIO(small_png_file.read())
     upload = MediaIoBaseUpload(fd=small_png_fd, mimetype='image/png',
@@ -641,7 +738,7 @@ class TestHttpRequest(unittest.TestCase):
     model = JsonModel()
 
     request = HttpRequest(
-        HttpMockWithSSLErrors(
+        HttpMockWithErrors(
             3, {'status': '200', 'location': 'location'}, '{"foo": "bar"}'),
         model.response,
         u'https://www.example.com/file_upload',
@@ -654,7 +751,10 @@ class TestHttpRequest(unittest.TestCase):
 
   def test_retry(self):
     num_retries = 5
-    resp_seq = [({'status': '500'}, '')] * num_retries
+    resp_seq = [({'status': '500'}, '')] * (num_retries - 3)
+    resp_seq.append(({'status': '403'}, RATE_LIMIT_EXCEEDED_RESPONSE))
+    resp_seq.append(({'status': '403'}, USER_RATE_LIMIT_EXCEEDED_RESPONSE))
+    resp_seq.append(({'status': '429'}, ''))
     resp_seq.append(({'status': '200'}, '{}'))
 
     http = HttpMockSequence(resp_seq)
@@ -679,6 +779,30 @@ class TestHttpRequest(unittest.TestCase):
     for retry_num in range(num_retries):
       self.assertEqual(10 * 2**(retry_num + 1), sleeptimes[retry_num])
 
+  def test_no_retry_succeeds(self):
+    num_retries = 5
+    resp_seq = [({'status': '200'}, '{}')] * (num_retries)
+
+    http = HttpMockSequence(resp_seq)
+    model = JsonModel()
+    uri = u'https://www.googleapis.com/someapi/v1/collection/?foo=bar'
+    method = u'POST'
+    request = HttpRequest(
+        http,
+        model.response,
+        uri,
+        method=method,
+        body=u'{}',
+        headers={'content-type': 'application/json'})
+
+    sleeptimes = []
+    request._sleep = lambda x: sleeptimes.append(x)
+    request._rand = lambda: 10
+
+    request.execute(num_retries=num_retries)
+
+    self.assertEqual(0, len(sleeptimes))
+
   def test_no_retry_fails_fast(self):
     http = HttpMockSequence([
         ({'status': '500'}, ''),
@@ -696,14 +820,80 @@ class TestHttpRequest(unittest.TestCase):
         headers={'content-type': 'application/json'})
 
     request._rand = lambda: 1.0
-    request._sleep = lambda _: self.fail('sleep should not have been called.')
+    request._sleep = mock.MagicMock()
 
-    try:
+    with self.assertRaises(HttpError):
       request.execute()
-      self.fail('Should have raised an exception.')
-    except HttpError:
-      pass
+    request._sleep.assert_not_called()
 
+  def test_no_retry_403_not_configured_fails_fast(self):
+    http = HttpMockSequence([
+        ({'status': '403'}, NOT_CONFIGURED_RESPONSE),
+        ({'status': '200'}, '{}')
+        ])
+    model = JsonModel()
+    uri = u'https://www.googleapis.com/someapi/v1/collection/?foo=bar'
+    method = u'POST'
+    request = HttpRequest(
+        http,
+        model.response,
+        uri,
+        method=method,
+        body=u'{}',
+        headers={'content-type': 'application/json'})
+
+    request._rand = lambda: 1.0
+    request._sleep =  mock.MagicMock()
+
+    with self.assertRaises(HttpError):
+      request.execute()
+    request._sleep.assert_not_called()
+
+  def test_no_retry_403_fails_fast(self):
+    http = HttpMockSequence([
+        ({'status': '403'}, ''),
+        ({'status': '200'}, '{}')
+        ])
+    model = JsonModel()
+    uri = u'https://www.googleapis.com/someapi/v1/collection/?foo=bar'
+    method = u'POST'
+    request = HttpRequest(
+        http,
+        model.response,
+        uri,
+        method=method,
+        body=u'{}',
+        headers={'content-type': 'application/json'})
+
+    request._rand = lambda: 1.0
+    request._sleep =  mock.MagicMock()
+
+    with self.assertRaises(HttpError):
+      request.execute()
+    request._sleep.assert_not_called()
+
+  def test_no_retry_401_fails_fast(self):
+    http = HttpMockSequence([
+        ({'status': '401'}, ''),
+        ({'status': '200'}, '{}')
+        ])
+    model = JsonModel()
+    uri = u'https://www.googleapis.com/someapi/v1/collection/?foo=bar'
+    method = u'POST'
+    request = HttpRequest(
+        http,
+        model.response,
+        uri,
+        method=method,
+        body=u'{}',
+        headers={'content-type': 'application/json'})
+
+    request._rand = lambda: 1.0
+    request._sleep =  mock.MagicMock()
+
+    with self.assertRaises(HttpError):
+      request.execute()
+    request._sleep.assert_not_called()
 
 class TestBatch(unittest.TestCase):
 
--- a/googleapiclient/__init__.py
+++ b/googleapiclient/__init__.py
@@ -14,3 +14,14 @@
 
 __version__ = "1.5.0"
 
+# Set default logging handler to avoid "No handler found" warnings.
+import logging
+
+try:  # Python 2.7+
+    from logging import NullHandler
+except ImportError:
+    class NullHandler(logging.Handler):
+        def emit(self, record):
+            pass
+
+logging.getLogger(__name__).addHandler(NullHandler())
--- a/googleapiclient/__init__.py
+++ b/googleapiclient/__init__.py
@@ -14,14 +14,3 @@
 
 __version__ = "1.5.0"
 
-# Set default logging handler to avoid "No handler found" warnings.
-import logging
-
-try:  # Python 2.7+
-    from logging import NullHandler
-except ImportError:
-    class NullHandler(logging.Handler):
-        def emit(self, record):
-            pass
-
-logging.getLogger(__name__).addHandler(NullHandler())
--- a/googleapiclient/discovery_cache/file_cache.py
+++ b/googleapiclient/discovery_cache/file_cache.py
@@ -29,7 +29,11 @@ import os
 import tempfile
 import threading
 
-from oauth2client.contrib.locked_file import LockedFile
+try:
+  from oauth2client.contrib.locked_file import LockedFile
+except ImportError:
+  # oauth2client < 2.0.0
+  from oauth2client.locked_file import LockedFile
 
 from . import base
 from ..discovery_cache import DISCOVERY_DOC_MAX_AGE
--- a/googleapiclient/discovery_cache/file_cache.py
+++ b/googleapiclient/discovery_cache/file_cache.py
@@ -29,11 +29,7 @@ import os
 import tempfile
 import threading
 
-try:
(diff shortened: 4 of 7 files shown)


googleapiclient/http.py
```diff
@@ -56,6 +56,10 @@
 from googleapiclient.model import JsonModel
 from oauth2client import util
 
+try:
```
jonparrott: does `six` not have an alias for this?

tests/test_http.py
```diff
@@ -704,6 +759,80 @@ def test_no_retry_fails_fast(self):
...
+        method=method,
+        body=u'{}',
+        headers={'content-type': 'application/json'})
+
+    request._rand = lambda: 1.0
+    request._sleep = lambda _: self.fail('sleep should not have been called.')
+
+    try:
+      request.execute()
+      self.fail('Should have raised an exception.')
```
jonparrott: please use `with self.raises()`.

tests/test_http.py
```diff
@@ -704,6 +759,80 @@ def test_no_retry_fails_fast(self):
...
+    request = HttpRequest(
+        http,
+        model.response,
+        uri,
+        method=method,
+        body=u'{}',
+        headers={'content-type': 'application/json'})
+
+    request._rand = lambda: 1.0
+    request._sleep = lambda _: self.fail('sleep should not have been called.')
```
jonparrott: please use a mock and `mock.assert_not_called()`. or `self.assertFalse(mock.called)`.

googleapiclient/http.py
```diff
@@ -91,12 +95,24 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
         continue
-    if resp.status < 500:
+    reason = ''
+    try:
+      if resp and resp.status == 403:
+        data = json.loads(content.decode('utf-8'))
+        reason = data['error']['errors'][0]['reason']
+        logging.warning('Encountered status %d with reason "%s"' %
+                        (resp.status, reason))
+    except Exception:
```
jonparrott: This is too broad. What exactly are you hoping to catch here?

googleapiclient/http.py
```diff
@@ -91,12 +95,24 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+      pass
+    if resp.status < 500 and not ((resp.status == 403 and
```
jonparrott: This is a bit difficult to follow. Could you break it into multiple if statements?

jonparrott:
Finished a review pass. Please let me know when you're ready for another.

tests/test_http.py
nathanielmanistaatgoogle: I think @jonparrott means `with self.assertRaises(<arguments>)`?

jonparrott: yes, sorry, used to `py.test` and everything melds together in my mind in the morning.

googleapiclient/http.py
eesheesh: @jonparrott not as far as I can tell from [this](http://pythonhosted.org/six/).

jonparrott: Fair enough. Is there any worry about catching socket errors that aren't timeouterrors that we should do different stuff with?

googleapiclient/http.py
eesheesh: @jonparrott The various cases where parsing failed, e.g. `content` wasn't populated, or `json.loads()` couldn't parse it, or it could parse it but `data['error']['errors'][0]['reason']` is an invalid path... And possibly other things I can't think of.

Is there a way to get a comprehensive list of exceptions that may be thrown? If not, I'd be risking failing the request due to a parsing error in the retry logic, which would be very bad... I'm fairly new to Python so I'd love to learn a better way! :)

googleapiclient/http.py
eesheesh: @jonparrott do you mean something like this?
```
    if resp.status < 500:
      if ((resp.status == 403 and
          reason in ['userRateLimitExceeded', 'rateLimitExceeded'])
          or resp.status == 429):
        continue
      else:
        break
```

googleapiclient/http.py
jonparrott: iirc `json.loads` only throws `ValueError`. In any case, you should scope the `try: except` as tightly as possible:

```python
if resp and resp.status == 403:
    try:
        data = json.loads(content.decode('utf-8'))
        reason = data['error']['errors'][0]['reason']
    except ValueError, KeyError:
        logging.warning('Invalid JSON content from response')
        reason = content.decode('utf-8')

    logging.warning('Encountered status %d with reason "%s"' %
        (resp.status, reason))
```

googleapiclient/http.py
jonparrott: yes

tests/test_http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: On Windows, in Python 2.7.11, a network timeout looks like this:
```
  File "C:\Python27_11\lib\site-packages\httplib2\__init__.py", line 1609, in request
    (response, content) = self._request(conn, authority, uri, request_uri, method, body, headers, redirections, cachekey)
  File "C:\Python27_11\lib\site-packages\httplib2\__init__.py", line 1351, in _request
    (response, content) = self._conn_request(conn, request_uri, method, body, headers)
  File "C:\Python27_11\lib\site-packages\httplib2\__init__.py", line 1272, in _conn_request
    conn.connect()
  File "C:\Python27_11\lib\site-packages\httplib2\__init__.py", line 928, in connect
    raise socket.error, msg
error: [Errno 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond
```

I tried catching `socket.timeout` and it doesn't catch the exception (as expected from this trace), so I believe it's better to retry on a non-retriable error than miss a retriable one - especially one as important as a network timeout.

googleapiclient/http.py
eesheesh: Done.

eesheesh:
Thanks Jon! All set (I hope) - please take another look.

googleapiclient/http.py
```diff
@@ -82,22 +86,45 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+    if resp and resp.status == 403 and content:
+      try:
+        data = json.loads(content.decode('utf-8'))
+        reason = data['error']['errors'][0]['reason']
+      except (ValueError, KeyError):
+        logging.warning('Invalid JSON content from response')
+        try:
+          reason = content.decode('utf-8')
+        except UnicodeDecodeError:
+          reason = str(content)
```
jonparrott: I'm unsure on this behavior. If the content can't be decoded as utf-8, it probably doesn't have a useful human-readable string representation. It's probably better to just pass here.

eesheesh: Done.

googleapiclient/http.py
jonparrott: No worries. See my comment below.

jonparrott:
@eesheesh Thank you for going through these rounds of review with me. This is a critical part of the library and it's important that we get it right.

This is not your fault, but re-reading over the `_retry_request` function I really don't like the way the logic is laid out. We should leave things better than we found them. Do you mind restructuring the function a little bit?

Something like this would be much easier to read:

```python
    resp = None
    content = None

    for retry_num in range(num_retries + 1):

        # Sleep before retrying.
        if retry_num > 0:
            sleep_time = rand() * 2 ** retry_num
            logging.warning(
                'Sleeping %.2f seconds before retry %d of %d for %s: %s %s%s' %
                (sleep_time, retry_num, num_retries, req_type, method, uri,
                ', following status: %d.' % resp.status if resp else ''))
            sleep(sleep_time)

        try:
          resp, content = http.request(uri, method, *args, **kwargs)

        # Retry on ssl errors and socket timeouts.
        except (ssl.SSLError, TimeoutError) as ex:
            # If this is a non-timeout socket error, re-raise it.
            if isinstance(ex, socket.error):
                if ex.errno != errno.ETIMEDOUT:
                    raise

            if retry_num == num_retries:
                raise
            else:
                logging.warning('Encountered exception %s.', repr(ex))
                continue

        # Retry 5xx errors.
        if resp.status >= 500:
            continue

        # Retry 429 errors
        if resp.status == 429:
            continue

        # For 403 errors, we have to check for the `reason` in the response to
        # determine if it is safe to retry.
        if resp.status == 403:
            if not content:
                break

            # Content is in JSON format.
            try:
                data = json.loads(content.decode('utf-8'))
                reason = data['error']['errors'][0]['reason']
            except (ValueError, KeyError):
                logging.warning('Invalid JSON content from response.')
                break

            logging.warning('Encountered status %d with reason "%s"' %
                (resp.status, reason))

            if reason not in ['userRateLimitExceeded', 'rateLimitExceeded']:
                break

    return resp, content

```

(disclaimer: I haven't tested this or anything, but in general this is how I would prefer the logic to flow)

eesheesh:
I'll work on these changes now, but just FYI I still won't make a distinction between different `socket.error` types... I did some testing on Python 2.7 across platforms and I don't think it's feasible to identify timeouts reliably. It would be safer to just retry every `socket.error`.

Windows: `socket.error: [Errno 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond`
Linux: `socket.error: [Errno 110] Connection timed out`
Mac: `socket.error: [Errno 60] Operation timed out`

eesheesh:
Done, PTAL. :)

googleapiclient/http.py
```diff
@@ -82,22 +88,55 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+      # If there's no details about the 403 type, don't retry.
+      if not content:
+        break
+
+      # Content is in JSON format.
+      try:
+        data = json.loads(content.decode('utf-8'))
+        reason = data['error']['errors'][0]['reason']
+      except (UnicodeDecodeError, ValueError, KeyError) as ex:
+        logging.warning('Invalid JSON content from response: %s', repr(ex))
```
jonparrott: I don't think it's necessary/useful to log the exception here.

jonparrott:
This is almost good by my standards

> it would be safer to just retry every socket.error

Is `exception.errno` consistent across the platforms (not the value, but it should match the same constant in the `errno` module).

eesheesh:
Ok, I'm now checking for the error code. I did leave one case where we retry if the error code isn't found in socket.errno, because reliability  is better than the slight performance gain of not retrying in this corner case.

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
     try:
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
+    # Retry on SSL errors and socket timeout errors.
+    except (ssl.SSLError, TimeoutError) as ex:
+      # If this is a socket.error, only retry if it's a timeout.
+      if isinstance(ex, socket.error) and not isinstance(ex, ssl.SSLError):
+        try:
+          # errno's contents differ by platform so we have to match by name.
+          if not (socket.errno.errorcode[ex.errno] in
```
jonparrott: How is this different from

```
import errno

if ex.errno == errno.ETIMEDOUT:

```

jonparrott:
Looks mostly good to me. Can you squash your commits. 

@nathanielmanistaatgoogle do you mind taking a final look at this?

(FYI: I'll be out of office until Thursday)

googleapiclient/http.py
eesheesh: The difference is that I need to check for `WSAETIMEDOUT` as well, but it isn't defined when not on Windows. I thought it's more readable and more elegant to check it like that instead of handling `AttributeError`.

eesheesh:
Ok, I *think* I squashed my commits (GitHub noob here), please let me know if I need to do anything else.

Thanks!

nathanielmanistaatgoogle:
Not quite squashed - take a look at the way this page says "@eesheesh wants to merge 12 commits into `google:master` from `eesheesh:master`" - there should be just one commit, not twelve.

eesheesh:
Sorry, missed a step. PTAL, thanks!

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
   resp = None
+  content = None
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
-      sleep(rand() * 2**retry_num)
+      # Sleep before retrying.
+      sleep_time = rand() * 2 ** retry_num
       logging.warning(
-          'Retry #%d for %s: %s %s%s' % (retry_num, req_type, method, uri,
+          'Sleeping %.2f seconds before retry %d of %d for %s: %s %s%s' %
```
nathanielmanistaatgoogle: Please don't nest one string interpolation inside another string interpolation?

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+          (sleep_time, retry_num, num_retries, req_type, method, uri,
           ', following status: %d' % resp.status if resp else ''))
+      sleep(sleep_time)
 
     try:
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
+    # Retry on SSL errors and socket timeout errors.
+    except (ssl.SSLError, TimeoutError) as ex:
```
nathanielmanistaatgoogle: Code of the form
```
except (<exception type 1>, <exception type 2>) as <name>:
  if isinstance(<name>, <exception type>):
    <statements>
```
invites the question of whether it can be of the form
```
except <exception type 1> as <name>:
  <statements>
except <exception type 2> as <possibly other name>:
  <statements>
```
. Does that question apply here? I'm having trouble seeing.

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+      # If this is a socket.error, only retry if it's a timeout.
+      if isinstance(ex, socket.error) and not isinstance(ex, ssl.SSLError):
+        try:
+          # errno's contents differ by platform so we have to match by name.
+          if not (socket.errno.errorcode[ex.errno] in
+              ['WSAETIMEDOUT', 'ETIMEDOUT']):
```
nathanielmanistaatgoogle: Never write a list literal where a tuple literal will do.

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+            raise
+        # If we couldn't match it - retry because we can't know for sure.
+        except KeyError:
+          pass
       if retry_num == num_retries:
         raise
       else:
+        logging.warning('Encountered exception %s', repr(ex))
```
nathanielmanistaatgoogle: What's the difference between `logging.warning('Encountered exception %s', repr(ex))` and `logging.warning('Encountered exception %r', ex)`? Why prefer the former?

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+    if resp.status >= 500:
+      continue
+
+    # Retry 429 errors.
+    if resp.status == 429:
+      continue
+
+    # For 403 errors, we have to check for the `reason` in the response to
+    # determine if it is safe to retry.
+    if resp.status == 403:
```
nathanielmanistaatgoogle: `six.moves.http_client.FORBIDDEN` rather than the literal integer value 403, please.

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
         continue
-    if resp.status < 500:
-      break
+
+    # Retry 5xx errors.
+    if resp.status >= 500:
+      continue
+
+    # Retry 429 errors.
+    if resp.status == 429:
```
nathanielmanistaatgoogle: Please define a module-scope constant for this value, probably named `_TOO_MANY_REQUESTS`.

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
 
     try:
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
+    # Retry on SSL errors and socket timeout errors.
+    except (ssl.SSLError, TimeoutError) as ex:
+      # If this is a socket.error, only retry if it's a timeout.
+      if isinstance(ex, socket.error) and not isinstance(ex, ssl.SSLError):
+        try:
+          # errno's contents differ by platform so we have to match by name.
```
nathanielmanistaatgoogle: The depth of code nesting here is pretty fearsome - what opportunity do we have to pull this logic out into a helper function?

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+
+      # Content is in JSON format.
+      try:
+        data = json.loads(content.decode('utf-8'))
+        reason = data['error']['errors'][0]['reason']
+      except (UnicodeDecodeError, ValueError, KeyError):
+        logging.warning('Invalid JSON content from response')
+        break
+
+      logging.warning('Encountered status 403 with reason "%s"' % reason)
```
nathanielmanistaatgoogle: `logging.warning(<format string>, <arguments to format string>)` rather than `logging.warning(<format string> % <arguments to format string>)`.

Fun fact: long ago in a different life I was the one who wrote the patch to pylint that inspects code for this issue.

googleapiclient/http.py
```diff
@@ -82,22 +89,65 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+
+      # Only retry rate limit related failures.
+      if reason not in ['userRateLimitExceeded', 'rateLimitExceeded']:
```
nathanielmanistaatgoogle: Never write a list literal when a tuple literal will do.

nathanielmanistaatgoogle:
(Completed first round of content review.)

eesheesh:
I think I addressed all of your comments. I found a way to catch both socket.error and TimeoutError that works on Python 2.7 and 3.4, but fails on Python 2.6 (see Travis results).

It appears that Python 2.6 is obsolete now, do we intend to keep supporting it? If we do, then I can define TimeoutError only when the Python version is 2.6, but it's an ugly hack, and considering that there's no real issue with Python 2.7, I don't think there's any point in filing an upstream bug.

What do you think?

tests/test_http.py
```diff
@@ -113,7 +116,46 @@ def request(self, *args, **kwargs):
...
+            # For Windows:
+            ex.errno = socket.errno.WSAETIMEDOUT
+          except AttributeError:
+            # For Linux/Mac:
+            ex.errno = socket.errno.ETIMEDOUT
+          # Now raise the correct timeout error.
+          raise ex
+        else:
+          raise TimeoutError()
+
```
nathanielmanistaatgoogle: Two blank lines between module-scope classes.

googleapiclient/http.py
```diff
@@ -81,23 +84,76 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
   Returns:
     resp, content - Response from the http request (may be HTTP 5xx).
   """
+
```
nathanielmanistaatgoogle: Drop this inserted blank line. Unless there's some reason for it that I'm missing?

googleapiclient/http.py
```diff
@@ -56,11 +58,12 @@
...
 
-
 DEFAULT_CHUNK_SIZE = 512*1024
 
 MAX_URI_LENGTH = 2048
 
+_TOO_MANY_REQUESTS = 429
+
 
 def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
```
nathanielmanistaatgoogle: With this change this function is now nearly one hundred lines long with seventy lines of body. What options do we have for breaking it up into an assembly of code elements each of which is a manageable size?

googleapiclient/http.py
```diff
@@ -81,23 +84,76 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+      # If there's no details about the 403 type, don't retry.
+      if not content:
+        break
+
+      # Content is in JSON format.
+      try:
+        data = json.loads(content.decode('utf-8'))
+        reason = data['error']['errors'][0]['reason']
+      except (UnicodeDecodeError, ValueError, KeyError):
+        logging.warning('Invalid JSON content from response')
```
nathanielmanistaatgoogle: Should we not include the invalid content in what gets logged?

googleapiclient/http.py
```diff
@@ -81,23 +84,76 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
     try:
+      exception = None
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
+    # Retry on SSL errors and socket timeout errors.
+    except ssl.SSLError as ex:
+      exception = ex
+    except socket.error as ex:
+      try:
+        # errno's contents differ by platfor,m so we have to match by name.
```
nathanielmanistaatgoogle: Typo "platfor,m".

googleapiclient/http.py
```diff
@@ -81,23 +84,76 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+        if not (socket.errno.errorcode[ex.errno] in
```
nathanielmanistaatgoogle: Prefer `<element> not in <sequence>` to `not <element> in <sequence>`.

googleapiclient/http.py
```diff
@@ -81,23 +84,76 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+            ('WSAETIMEDOUT', 'ETIMEDOUT')):
+          raise
+      # If we couldn't match it - retry because we can't know for sure.
+      except KeyError:
+        pass
+      exception = ex
+    except TimeoutError as ex:
```
nathanielmanistaatgoogle: In what environment (platform, Python version, and so on) is a `TimeoutError` not also a `socket.error`?

eesheesh: Oh, I didn't know `TimeoutError` is also a `socket.error`! Changed accordingly.

googleapiclient/http.py
eesheesh: I could move the logic to decide what to do based on `resp.status` and `content`, but I think this would be less readable overall. If you think this is better I'm happy to do it.

eesheesh:
Thanks for the review, I addressed the comments, please let me know if you want me to break the method down.

googleapiclient/http.py
nathanielmanistaatgoogle: Well, I only just found out myself, so don't take my word for it; make your own independent verification.

googleapiclient/http.py
nathanielmanistaatgoogle: Yes please. A seventy-line method body is fundamentally unreadable.

eesheesh:
Ok, broken into two methods, and rebased the branch to avoid conflicts.

googleapiclient/http.py
```diff
@@ -84,26 +88,90 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+      exception = None
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
+    # Retry on SSL errors and socket timeout errors.
+    except ssl.SSLError as ex:
+      exception = ex
+    except socket.error as ex:
+      try:
+        # errno's contents differ by platform, so we have to match by name.
+        if (socket.errno.errorcode[ex.errno] not in
```
nathanielmanistaatgoogle: Rather than nest a `try` inside an `except`, how about writing `socket.errno.errorcode.get(ex.errno, None)`?

googleapiclient/http.py
```diff
@@ -84,26 +88,90 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
         continue
-    if resp.status < 500:
+
+    if not _should_retry_response(resp.status, content):
       break
 
   return resp, content
 
 
+def _should_retry_response(resp_status, content):
```
nathanielmanistaatgoogle: Please place this function earlier in the file ("above") any of its use-sites.

googleapiclient/http.py
```diff
@@ -84,26 +88,90 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+  """Determines whether a response justifies a retry.
+
+  Args:
+    resp_status: The response status received.
+    content: The response content body. 
+
+  Returns:
+    True if the response needs to be retried, otherwise False.
```
nathanielmanistaatgoogle: In the space of ten lines you equate between "should", "justifies", and "needs" - pick one? Probably "should".

googleapiclient/http.py
```diff
@@ -84,26 +88,90 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+    except socket.error as ex:
+      try:
+        # errno's contents differ by platform, so we have to match by name.
+        if (socket.errno.errorcode[ex.errno] not in
+            ('WSAETIMEDOUT', 'ETIMEDOUT')):
+          raise
+      # If we couldn't match it - retry because we can't know for sure.
+      except KeyError:
+        pass
+      exception = ex
```
nathanielmanistaatgoogle: What's the difference between `exception` and `ex`? Why should this method have both fields? Why doesn't one suffice?

googleapiclient/http.py
```diff
@@ -84,26 +88,90 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+  exception = None
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
-      sleep(rand() * 2**retry_num)
-      LOGGER.warning(
-          'Retry #%d for %s: %s %s%s' % (retry_num, req_type, method, uri,
-          ', following status: %d' % resp.status if resp else ''))
+      # Sleep before retrying.
+      sleep_time = rand() * 2 ** retry_num
+      LOGGER.warning('Sleeping %.2f seconds before retry %d of %d for %s: '
```
nathanielmanistaatgoogle: If you line-wrap after the opening parenthesis, can you avoid the implicit string concatenation? Always avoid implicit string concatenation whenever you can.

googleapiclient/http.py
eesheesh: Because of the scope. I'm using `exception` in the logged message. I verified that the caught exception would be out of scope:
```python
>>> test = None
>>> try:
...   raise TimeoutError()
... except Exception as test:
...   pass
... 
>>> test
  File "<stdin>", line 1, in <module>
NameError: name 'test' is not defined
```

nathanielmanistaatgoogle: Wow. [Today](http://stackoverflow.com/questions/29268892/python-3-exception-deletes-variable-in-enclosing-scope-for-unknown-reason) [I](https://www.python.org/dev/peps/pep-3110/) [learned](https://www.python.org/dev/peps/pep-0344/).

Does `exception` need to be assigned to `None` on both line 92 and line 104?

Since there can't be exactly one field, how about more specific names than `ex` for the shortest-lifetime fields? What would you think of `except ssl.SSLError as ssl_error:` and `except socket.error as socket_error`?

eesheesh: Changes made. Thanks!

nathanielmanistaatgoogle:
@eesheesh: there are still a lot of outstanding code review comments - are you seeing everything I've posted? GitHub code reviews leave a lot to be desired...

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

tests/test_http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

googleapiclient/http.py
eesheesh: Done.

eesheesh:
@nathanielmanistaatgoogle Sorry for not being more verbose - I replied "Done." now to all of the comments I received and addressed over the last 14 days (since first squishing commits). Please let me know if there's still anything I missed.

Thanks!

googleapiclient/http.py
```diff
@@ -84,21 +131,37 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
-      if retry_num == num_retries:
+    # Retry on SSL errors and socket timeout errors.
+    except ssl.SSLError as ssl_error:
+      exception = ssl_error
+    except socket.error as socket_error:
+      # errno's contents differ by platform, so we have to match by name.
+      if (socket.errno.errorcode.get(socket_error.errno) not in
+          ('WSAETIMEDOUT', 'ETIMEDOUT')):
```
nathanielmanistaatgoogle: By placing the opening parenthesis of the tuple on the line above you can spare yourself the outer parentheses. I'm under the impression that that's what's considered idiomatic form, but when I try to think about it deeply I just feel embarrassed for the parser.

Someday an auto-formatter will run over this code.

tests/test_http.py
```diff
@@ -113,7 +116,45 @@ def request(self, *args, **kwargs):
       return httplib2.Response(self.success_json), self.success_data
     else:
       self.num_errors -= 1
-      raise ssl.SSLError()
+      if self.num_errors == 1:
+        raise ssl.SSLError()
+      else:
+        if sys.version_info[0] == 3:
```
nathanielmanistaatgoogle: `six.PY3` is already imported into this module and is the general way that we determine twoness and threeness throughout this codebase. Any reason why this location should be different?

googleapiclient/http.py
eesheesh: Done.

tests/test_http.py
eesheesh: Done.

googleapiclient/http.py
```diff
@@ -84,21 +131,37 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
...
+      exception = None
       resp, content = http.request(uri, method, *args, **kwargs)
-    except ssl.SSLError:
-      if retry_num == num_retries:
+    # Retry on SSL errors and socket timeout errors.
+    except ssl.SSLError as ssl_error:
+      exception = ssl_error
+    except socket.error as socket_error:
+      # errno's contents differ by platform, so we have to match by name.
+      if socket.errno.errorcode.get(socket_error.errno) not in ('WSAETIMEDOUT',
```
nathanielmanistaatgoogle: Line-wrap _immediately_ after the opening parenthesis. Both elements of the tuple should go on the following line. And while we're dwelling on minutia: add a comma after the last tuple element.

eesheesh: Done.

nathanielmanistaatgoogle:
GitHub is not yet showing me squashed commits.

@jonparrott: please rereview? I asked for, and got, many changes. Feel free to tell me which were mistakes on my part.

eesheesh:
I haven't squashed commits again yet, waiting to see if you have any additional feedback. Let me know when you want me to squash.

googleapiclient/http.py
```diff
@@ -63,6 +65,51 @@
...
+  if resp_status == six.moves.http_client.FORBIDDEN:
+    # If there's no details about the 403 type, don't retry.
+    if not content:
+      return False
+
+    # Content is in JSON format.
+    try:
+      data = json.loads(content.decode('utf-8'))
+      reason = data['error']['errors'][0]['reason']
+    except (UnicodeDecodeError, ValueError, KeyError):
```
jonparrott: (no action required): I'd prefer to see a separate catch for KeyError, as it doesn't indicate invalid JSON, just that the JSON wasn't in the format you expected.

jonparrott:
This LGTM, and I'm strongly in favor of the separate `should_retry` function - perhaps in the future we could even let users override that.

@eesheesh thank you so much for doing this and sticking with it through our detailed review.

eesheesh:
<off topic but...> I updated my branch to resolve conflicts and now I see other people's commits there, and googlebot thinks I don't have a CLA for them.

Can you suggest a safe way to squash the commits on https://github.com/eesheesh/google-api-python-client/commits/master to resolve this?

jonparrott:
Looks like some weirdness has occurred with your branch?

Make a backup of your current master, just in case:

   git checkout -b master-backup

Pull upstream (assuming you have google/google-api-python-client as upstream):

   git fetch upstream master

Then maybe try an an interactive rebase to fix your commits on top of master?

   git rebase -i upstream/master

jonparrott:
If that fails, I can manually merge this.

eesheesh:
That gives me all of these commits:
```
pick 69296f9 Default to octet-stream if mimetype detection fails.
pick 64f0a40 Handle SSL errors with retries
pick 74dfe73 Make compatible with oauth2client v2.0.0
pick 4fffd9c Specify allowed version range for dependencies
pick 88961e5 Fix issue where docs would fail to build if an API is whitelisted.
pick 9e3bc6c Bump version to 1.5.0
pick 7949171 Updated docs
pick e2fb55e Updating CHANGELOG
pick a4b8370 pep8 + print statement alignment
pick ca60f1f Re-adjusted the spacing in the comments.
pick 2a276c6 Make compatible with oauth2client v2.0.0 xsrfutil.py
pick 74a96e7 Fallback to use new discovery uri pattern when the old one fails.
pick 76010c9 move logging into __name__-specific loggers from the root logger
pick 186cb5d Add retry on rate limiting API responses and network timeouts
pick 8848f3e Revert "Add retry on rate limiting API responses and network timeouts"
pick f836a95 Revert "Add retry on rate limiting API responses and network timeouts"
pick c03ff89 Fix merge
pick 8350677 Fix whitespace
```
Should I just squash all of them?

jonparrott:
Generally, you can just move all yours to the end, and squash all of yours. Please don't squash someone else's commit into yours.

(If you're still having trouble, I'm happy to just manually merge this).

eesheesh:
Ok, I think it's ready now. Let me know if you need any other actions from my side. Thanks!

jonparrott:
This is weird, it's still showing 8 commits?

eesheesh:
7 of those aren't mine...

I have backups of my changes, is there a way to reset the branch to be exactly like the upstream, and then reapply the changes?

Alternatively, I can try to drop these commits altogether (in the interactive rebase) and see what happens...

jonparrott:
I'll just manually merge. Stand by.

jonparrott:
I've cherry-picked and merged this as c6425a00e. Thank you so much for your contribution.

eesheesh:
Excellent, thanks!

PR #178, opened 2016-02-12
Sample Python app for the new Analytics Reporting API V4

ikuleshov:
This is just a preview and should not be merged anywhere yet.


samples/analytics/core_reporting_v4_reference.py
```diff
@@ -0,0 +1,600 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2014 Google Inc. All Rights Reserved.
```
nathanielmanistaatgoogle: My calendar says that the current year is 2016.

PR #180, opened 2016-02-16
Fix incompatibility with oauth2client v2.0.0

adrian-the-git:
LockedFile moved to contrib subpackage


adrian-the-git:
I signed it!

PR #181, opened 2016-02-16
Fix incompatibility with oauth2client v2.0.0

adrian-the-git:
Several bits moved to contrib subpackage of oauth2client


nathanielmanistaatgoogle:
Closed? What happened?

adrian-the-git:
There was just more to it

PR #185, opened 2016-02-18
Import LockedFile from contrib to work with latest oauth2client

thejens:
It seems the new oauth2client has moved locked file into contrib, and hence my applications throw a warning that it failed to import the module.


thejens:
Realized this is taken care of in open PR https://github.com/google/google-api-python-client/pull/182 , closing this one

PR #189, opened 2016-02-20
Release v1.5.0

jonparrott:
@nathanielmanistaatgoogle @tmatsuo PTAL.


tmatsuo:
LGTM except for one comment

jonparrott:
> This commit looks behaviorally interesting and non-trivial

A few things:
* It's pretty trivial. There are some APIs that are returned by the discovery endpoint that do not yet have published discovery docs. This will try to build those APIs anyway and explode. The APIs that cause this aren't public yet and therefore should not be part of the public documentation.
* It does not change behavior; if anything, it ensures that the current behavior is maintained despite a change in expectations from the discovery service.
* This change is in our docs building tool and not the library itself.

jonparrott:
> Please maintain no-blank-line-between-doc-string-and-behavior-body in this codebase?

Fixed.

nathanielmanistaatgoogle:
Looks good and thanks for the release. :-)

jonparrott:
Thanks, @nathanielmanistaatgoogle. :)

jonparrott:
Published to [PyPI](https://pypi.python.org/pypi/google-api-python-client). #190 has been opened to both automate and document this process in the future.

PR #192, opened 2016-02-23
pep8 + print statement alignment

danielquinn:
The original:

* Violated pep8
* Used `print()` both as a statement and as a function, guaranteeing that the code won't work in either Python 2.5 or 3.x
* Didn't explain anything about what was happening.

This change fixes two of those problems.


danielquinn:
I signed it!

jonparrott:
LGTM, thank you for your contribution!

PR #201, opened 2016-03-14
Allow using oauth2client versions < 2

toabctl:
Be compatible with older oauth2client versions. This simplifies
the installation (also for downstream) when software needs an older
oauth2client version but also wants to use googleapiclient.

### DIFF
--- a/googleapiclient/discovery_cache/file_cache.py
+++ b/googleapiclient/discovery_cache/file_cache.py
@@ -29,7 +29,11 @@ import os
 import tempfile
 import threading
 
-from oauth2client.contrib.locked_file import LockedFile
+try:
+  from oauth2client.contrib.locked_file import LockedFile
+except ImportError:
+  # oauth2client < 2.0.0
+  from oauth2client.locked_file import LockedFile
 
 from . import base
 from ..discovery_cache import DISCOVERY_DOC_MAX_AGE
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ packages = [
 
 install_requires = [
     'httplib2>=0.8,<1',
-    'oauth2client>=2.0.0,<3',
+    'oauth2client',
     'six>=1.6.1,<2',
     'uritemplate>=0.6,<1',
 ]
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,11 @@
 [tox]
-envlist = py26, py27, py33, py34
+envlist = py{26,27,33,34}-oauth2client{1,2}
 
 [testenv]
-deps = keyring
+deps =
+       oauth2client1: oauth2client<2
+       oauth2client2: oauth2client>=2,<=3
+       keyring
        mox
        pyopenssl
        pycrypto==2.6


jonparrott:
Are you sure this is the only change needed?

toabctl:
```I signed it!```

jonparrott:
@toabctl I'm also curious about what cases you're running into where you need a lower version of `oauth2client`.

toabctl:
> Are you sure this is the only change needed?

No. But looking at the changelog doesn't indicate any other needed changes and also your testsuite passes :)

jonparrott:
> your testsuite passes :)

That gives me a lot of confidence.

@nathanielmanistaatgoogle any opposition to this change?

toabctl:
> @toabctl I'm also curious about what cases you're running into where you need a lower version of oauth2client.

@jonparrott ```google-api-python-client``` needs ```oauth2client >= 2``` but the google cloud sdk (https://cloud.google.com/sdk/ ) still uses oauth2client 1.X . I know it's bundled in google-cloud-sdk but  distributions (I'm working for SUSE, but same for Debian, ...) bundled software is usually split so if there's a (security) fix you don't have to patch the software (in this case oauth2client) 2x. So if I want to package (RPM) and distribute google-api-python-client together with google-cloud-sdk, that's currently not possible.

jonparrott:
@toabctl interesting, we seem to get a lot of stuff around OS-level packages. Why not use a virtualenv?

/cc @dhermes

toabctl:
And I havn't checked any other software yet. There might be more which depends on oauth2client < 2

toabctl:
> @toabctl interesting, we seem to get a lot of stuff around OS-level packages. Why not use a virtualenv?

Because as a user I want to have a single package manager to handle all my files. virtualenvs are nice to play and develop but not so nice for production. If I use a couple of different languages on my system, should I use 10 different package managers to handle all my files (gem, pip, npm, ...)?

toabctl:
So *if* you merge this it would be really cool to get a new release on pypi btw.

jonparrott:
>  virtualenvs are nice to play and develop but not so nice for production. [...]

I could definitely disagree with this sentiment, but this is neither the time nor place.

Will wait for @nathanielmanistaatgoogle's opinion on this change before merging.

toabctl:
@jonparrott but even with a virtualenv it can happen that you need to combine 2 python modules where module 1 needs oauth2client < 2 and module 2 needs oauth2client >=2

nathanielmanistaatgoogle:
This is weird and strange, so the barrier to entry is going to be high.

(1) Thank you for the test coverage.

(2) Thank you for giving a specific example with real libraries explaining why this should exist.

(3) How long would we want to live with this? Would this have an expiration date after which we could remove it and again just depend on oauth2client>=2.0.0?

toabctl:
> How long would we want to live with this? Would this have an expiration date after which we could remove it and again just depend on oauth2client>=2.0.0?

As long as possible. But at least as long as other software from google uses still an old oauth2client :)

jonparrott:
@toabctl what software is still using old oauth2client?

jonparrott:
I see above that the cloud sdk does, but it normally vendors its dependencies. Does the debian package actually depend on a separate oauth2client package?

toabctl:
> I see above that the cloud sdk does, but it normally vendors its dependencies. Does the debian package actually depend on a separate oauth2client package?

The openSUSE package does (but Debian is doing the same - removing included 3rd party modules from a package). See https://build.opensuse.org/package/show/openSUSE:Factory/google-cloud-sdk

jonparrott:
Thank you for that information. I'm fine to merge if @nathanielmanistaatgoogle is.

PR #202, opened 2016-03-14

ethanbao:

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -82,6 +82,9 @@ URITEMPLATE = re.compile('{[^}]*}')
 VARNAME = re.compile('[a-zA-Z0-9_-]+')
 DISCOVERY_URI = ('https://www.googleapis.com/discovery/v1/apis/'
                  '{api}/{apiVersion}/rest')
+V1_DISCOVERY_URI = DISCOVERY_URI
+V2_DISCOVERY_URI = ('https://{api}.googleapis.com/$discovery/rest?'
+                    'version={apiVersion}')
 DEFAULT_METHOD_DOC = 'A description of how to use this function'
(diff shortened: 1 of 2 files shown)


nathanielmanistaatgoogle:
Looks good; @jonparrott would you like to be a second set of eyes on this?

googleapiclient/discovery.py
```diff
@@ -82,6 +82,9 @@
 VARNAME = re.compile('[a-zA-Z0-9_-]+')
 DISCOVERY_URI = ('https://www.googleapis.com/discovery/v1/apis/'
```
jonparrott: Is this constant used anymore? If not, can we remove it?

jonparrott:
LGTM with one small question.

googleapiclient/discovery.py
nathanielmanistaatgoogle: It's a part of the public API, so we can't remove it until we bump to 2.0.0, regardless of the fact that we ourselves are no longer using it.

Always strive for a minimal public API in a software library. :-\

ethanbao: On Tue, Mar 15, 2016 at 6:59 PM, Nathaniel Manista <notifications@github.com
> wrote:

> In googleapiclient/discovery.py
> <https://github.com/google/google-api-python-client/pull/202#discussion_r56272232>
> :
>
> > @@ -82,6 +82,9 @@
> >  VARNAME = re.compile('[a-zA-Z0-9_-]+')
> >  DISCOVERY_URI = ('https://www.googleapis.com/discovery/v1/apis/'
>
> It's a part of the public API, so we can't remove it until we bump to
> 2.0.0, regardless of the fact that we ourselves are no longer using it.
>
> Always strive for a minimal public API in a software library. :-\
>

Yes. Thanks for the review.

jonparrott: Fair enough. :/

PR #220, opened 2016-04-20
googleapiclient.http: guard when importing ssl

### LINKED ISSUE #191: ImportError: No module named _ssl on App Engine since 1.5.0
After updating to 1.5.0, my App running on App Engine stopped working:

```
  File "<...>/base_api_client.py", line 4, in <module>
  File "<...>/lib/googleapiclient/discovery.py", line 63, in <module>
    from googleapiclient.http import BatchHttpRequest
  File "<...>/lib/googleapiclient/http.py", line 39, in <module>
    import ssl
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py", line 97, in <module>
    import _ssl             # if we can't import it, let the error propagate
  File "/Applications/GoogleAppEngineLauncher.app/Contents/Resources/GoogleAppEngine-default.bundle/Contents/Resources/google_appengine/google/appengine/tools/devappserver2/python/sandbox.py", line 948, in load_module
    raise ImportError('No module named %s' % fullname)
ImportError: No module named _ssl
```

The error goes away if I add the ssl module to my app.yaml:

```
- name: ssl
  version: latest
```

However, googleapiclient worked fine previously without using/importing the ssl module, so I am not sure this is intended.

Thanks,

rctay:
Fix #191 by adding a guard when importing `ssl`, and replace all direct
references to `SSLError` (the sole member of `ssl` being used) with a
shim, `_ssl_SSLError`. It is prefixed with a '_' to caution users
against relying on the name when importing `googleclient.http`.

`httplib2` also takes the same approach (see `ssl_SSLError` in
`httplib2/__init__.py`), but instead of using `None` when faking
`SSLError`, we provide a 'real' class, so that code can continue to
trigger the exception handler for `SSLError` in `googleclient.http` (eg.
in tests, `SSLError`'s are raised). If the shim `_ssl_SSLError` were
`None`, an exception handler for it would not catch anything, as `None`
is not a class and we could never throw anything that could be
"compatible" with `None` [1]

[1] https://docs.python.org/2/reference/compound_stmts.html#except

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -38,10 +38,17 @@ import mimetypes
 import os
 import random
 import socket
-import ssl
 import sys
 import time
 import uuid
+
+# TODO(issue 221): Remove this conditional import jibbajabba.
+try:
+  import ssl
+except ImportError:
+  _ssl_SSLError = object()
+else:
+  _ssl_SSLError = ssl.SSLError
 
 from email.generator import Generator
 from email.mime.multipart import MIMEMultipart
@@ -146,7 +153,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
       exception = None
       resp, content = http.request(uri, method, *args, **kwargs)
     # Retry on SSL errors and socket timeout errors.
-    except ssl.SSLError as ssl_error:
+    except _ssl_SSLError as ssl_error:
       exception = ssl_error
     except socket.error as socket_error:
       # errno's contents differ by platform, so we have to match by name.


jonparrott:
What situations are you in where `ssl` isn't available? It's part of the standard library since 2.6.

jonparrott:
Ah okay, I see, I read the issue.

So the issue is that App Engine doesn't have the SSL module available by default.

jonparrott:
@nathanielmanistaatgoogle I am on the fence about App Engine-specific changes. WDYT?

rctay:
@jonparrott Thanks for looking. Perhaps we can continue the thread at #191?

jonparrott:
Nah, if @nathanielmanistaatgoogle is cool with this we can merge. :)

rctay:
By the way, I have ran the tests with tox for py27-oauth2client1/2 and they remain working. I also tested on App Engine, by listing a Cloud Storage bucket:

```
from google.appengine.api import app_identity
from oauth2client.contrib.appengine import AppAssertionCredentials

import httplib2
import webapp2

class MyHandler(webapp2.ResponseHandler):
  def get(self):
    credentials = AppAssertionCredentials('https://www.googleapis.com/auth/devstorage.read_write')
    http_auth = credentials.authorize(httplib2.Http())
    service = build('storage', 'v1', http=http_auth)
    res = service.objects().list(bucket=app_identity.get_default_gcs_bucket_name()).execute()
    self.response.write(repr(res))
```

googleapiclient/http.py
```diff
@@ -38,11 +38,26 @@
...
+  import ssl
+except ImportError:
+  class _ssl_SSLError(socket.error):
+    # To be precise, ssl.SSLError only had `library` and `reason` since Python 2.7.9,
+    # but the primary purpose of this shim is not for compatibility between
+    # Python versions, but with environments that don't have `ssl`, like App
+    # Engine.
+    library = ''
+    reason = ''
+
```
nathanielmanistaatgoogle: Drop this blank line?

googleapiclient/http.py
```diff
@@ -38,11 +38,26 @@
...
 import sys
 import time
 import uuid
 
+
+try:
+  import ssl
+except ImportError:
+  class _ssl_SSLError(socket.error):
+    # To be precise, ssl.SSLError only had `library` and `reason` since Python 2.7.9,
```
nathanielmanistaatgoogle: Why attempt compatibility? Why not have the class loudly advertise that it is *not* an `ssl.SSLError`, because the code is executing in a place in which the standard library `ssl` module is not available? If someone has a problem that proceeds as far as handling this exception, why not be as helpful to them as we can be?

nathanielmanistaatgoogle:
@jonparrott: [My pride is wounded as advertised](https://github.com/google/google-api-python-client/issues/191#issuecomment-187978863).

jonparrott:
@nathanielmanistaatgoogle FWIW ssl by default should hopefully happen this quarter, but no promises.

jonparrott:
With that in mind I agree the absence of the ssl module should not have compatibility. ;)

rctay:
@nathanielmanistaatgoogle Sorry to have wounded it. Dropped the blank line and make accesses to `library` and `reason` fail. Hopefully it's loud enough and hope this makes your wound hurt less!

tests/test_http.py
```diff
@@ -116,7 +116,7 @@ def request(self, *args, **kwargs):
     else:
       self.num_errors -= 1
       if self.num_errors == 1:
-        raise ssl.SSLError()
```
nathanielmanistaatgoogle: I'm not sure about the change to this file. If we're not going to set up infrastructure that runs our tests without the `ssl` module available, changing the test code is moot, isn't it?

googleapiclient/http.py
```diff
@@ -38,11 +38,34 @@
...
-import ssl
 import sys
 import time
 import uuid
 
+
+try:
+  import ssl
+except ImportError:
+  class _ssl_SSLError(socket.error):
```
nathanielmanistaatgoogle: I don't see why this class is necessary - it's never instantiated, so the properties defined within it can never be accessed, right? The problem seems to be that _if_ the `ssl` module is available, we want to be able to catch instances of `ssl.SSLError`, but that if the `ssl` module is not available, _we only want the code not to crash_. Right?

Why isn't the code that you're adding to the imports of this module just
```
try:
  import ssl
except ImportError:
  _ssl_SSLError = object()
else:
  _ssl_SSLError = ssl.SSLError
```
?

rctay:
You're right, we don't run tests on App Engine, or any environment without `ssl` for that matter. I have dropped the changes to the test infrastructure, as well as omit implementing `SSLError`.

nathanielmanistaatgoogle:
@jonparrott: please take another look at this? I feel like where I've directed this change is correct, but also pretty radical. Any qualms? Also, any desire for a comment in the clearly weird code? Should we open an issue for removing this as soon as possible?

jonparrott:
@nathanielmanistaatgoogle This LGTM and I'm okay with it. I agree we should file a bug to remove this once ssl is on by default in GAE.

googleapiclient/http.py
```diff
@@ -38,11 +38,19 @@
 import os
 import random
 import socket
-import ssl
 import sys
 import time
 import uuid
 
+
+try:
```
nathanielmanistaatgoogle: Please add `TODO(issue 221): Remove this conditional import jibbajabba.` immediately above this line.

googleapiclient/http.py
```diff
@@ -38,11 +38,19 @@
...
+  import ssl
+except ImportError:
+  _ssl_SSLError = object()
+else:
+  _ssl_SSLError = ssl.SSLError
+
```
nathanielmanistaatgoogle: Single blank lines, not double blank lines, above and below this entire added section. It's special, but it's not double-blank-lines special.

rctay:
Done!

PR #223, opened 2016-05-03
Changed type in docs for HttpRequest

antsankov:
See https://github.com/google/google-api-python-client/issues/222

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -742,7 +742,7 @@ class HttpRequest(object):
                 on an error.
       uri: string, the absolute URI to send the request to
       method: string, the HTTP method to use
-      body: string, the request body of the HTTP request,
+      body: dict, the request body of the HTTP request,
       headers: dict, the HTTP request headers
       methodId: string, a unique identifier for the API method being called.
       resumable: MediaUpload, None if this is not a resumbale request.


antsankov:
Already a Googler (tsankov@google.com).

jonparrott:
If you're a googler, please register your github username over at `go/github`. 

So, this is weird, but a few lines down it treats body as a string. Can you share some code that reproduces the issue you saw?

antsankov:
@jonparrott: 

I discovered this issue when I was trying to use the [predict API](https://cloud.google.com/prediction/docs/). 

My request looked like this: 
```
  request_body = {'input': {'csvInstance': [text_input]}}

  prediction_request = service.trainedmodels().predict(
      project=dg.PROJECT_ID, id=model, body=request_body)
```

It would say my input was invalid until I left `request_body` as a dict.

jonparrott:
Hmm, I'm failing on finding where this code treats it like a dict.

jonparrott:
@antsankov ping?

antsankov:
Hello @jonparrott 

Summary: I suspect the problem lies in the encoding of the body for the request. While the eventual http request needs to be a string, it needs to go through a conversion before it can be pushed to the http2 lib for the eventual call. The post works when we give it a dictionary because there is a json.dumps call in between these two parts. 

This mismatch is mitigated by the [to_json](https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L347) method which is really just a wrapper around json.dumps. The test cases use an empty json ```'{}'``` for body ([1](https://github.com/google/google-api-python-client/blob/master/tests/test_http.py#L246), [2](https://github.com/google/google-api-python-client/blob/master/tests/test_http.py#L735)) and this doesn't properly expose this encoding issue. Which is that *json.dumps doesn't treat a conversion of a string, the same as it treats the conversion of a dict.* 

As a side note, I no longer work for google, so I don't have access to my predict API code that exposes this error, however @zdv-1993 comment shows something similar along with my example and the stackoverflow post reference in the issue. All of these work with dicts and not strings because, I suspect, that is the format needed by json.dumps to eventually pass on a valid request. 

If this evidence still isn't sufficient I would propose either authoring more test cases, or investigating how the encoding changes when the execute method is called up the stack to when it reaches the HTTP2 lib call.

jonparrott:
@antsankov I'd love to see some failing test cases for this.

chainoy:
Hi

PR #239, opened 2016-06-15
TEMPLATE_DIRS updated

kunal95:


nathanielmanistaatgoogle:
We can't integrate your contribution until you've signed the CLA.

samples/django_sample/settings.py
```diff
@@ -66,12 +66,12 @@
 
 ROOT_URLCONF = 'django_sample.urls'
 
-TEMPLATE_DIRS = (
+TEMPLATE_DIRS = [
     # Put strings here, like "/home/html/django_templates"
     # Always use forward slashes, even on Windows.
     # Don't forget to use absolute paths, not relative paths.
     os.path.join(os.path.dirname(__file__), 'templates')
```
nathanielmanistaatgoogle: Rather than change `TEMPLATE_DIRS` to a list, shouldn't a comma simply be added to the end of its one element?

kunal95: No, there would still be Type error. It expects a list.

jonparrott:
@waprin can you take a look at this simple change?

waprin:
Looks like TEMPLATE_DIRS was changed from tuple to list in 1.9, but the field has also been deprecated since 1.8 so I think any change should just get rid of it. The whole sample is so dated that it needs other updates to even run. I have it on my plate to revisit all these samples, @kunal95 if you want to fix it in the meantime that's fine, but I think the PR might as well totally fix the sample for 1.8 and 1.9 with no deprecated features used.

kunal95:
Yes, seems like now they have DIRS option within TEMPLATES. I hope I didn't add anything new to your plate.

jonparrott:
Closing this in favor of @waprin taking another look.

PR #261, opened 2016-08-09
Add a warning when using mocking without cache.

ternus:
See https://github.com/google/google-api-python-client/issues/208 .
If HttpMock or HttpMockSequence is used in a test without the cache
being filled first, a strange KeyError will be thrown. This happens
on e.g. Travis CI. Unfortunately this behavior is used in the tests;
this patch adds a warning for those unfortunate enough to encounter
it.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -315,6 +315,15 @@ def build_from_document(
 
   if isinstance(service, six.string_types):
     service = json.loads(service)
+
+  if  'rootUrl' not in service and (isinstance(http, (HttpMock,
+                                                      HttpMockSequence))):
+      logger.error("You are using HttpMock or HttpMockSequence without" +
+                   "having the service discovery doc in cache. Try calling " +
+                   "build() without mocking once first to populate the " +
+                   "cache.")
+      raise InvalidJsonError()
+
   base = urljoin(service['rootUrl'], service['servicePath'])
   schema = Schemas(service)
 


ternus:
Signed.

googleapiclient/discovery.py
```diff
@@ -315,6 +315,15 @@ def build_from_document(
 
   if isinstance(service, six.string_types):
     service = json.loads(service)
+
+  if not 'rootUrl' in service and (isinstance(http, HttpMock) or
```
nathanielmanistaatgoogle: Favor the more idiomatic `'rootUrl' not in service`.

googleapiclient/discovery.py
```diff
@@ -315,6 +315,15 @@ def build_from_document(
 
   if isinstance(service, six.string_types):
     service = json.loads(service)
+
+  if  'rootUrl' not in service and (isinstance(http, HttpMock) or
```
nathanielmanistaatgoogle: No need for two `isinstance` calls; [you can just pass both classes to a single call](https://docs.python.org/2/library/functions.html#isinstance):
```
isinstance(http, (HttpMock, HttpMockSequence,))
```
.

nathanielmanistaatgoogle:
Sorry to have missed that on first review. Please also squash commits and follow [these guidelines](http://chris.beams.io/posts/git-commit/#seven-rules) in your commit message.

ternus:
@nathanielmanistaatgoogle Squashed and reworded commit message. Does everything look OK?

nathanielmanistaatgoogle:
I'm satisfied. @jonparrott will also review and will merge.

jonparrott:
LGTM, thank you!

PR #268, opened 2016-08-20
Added requirements file to Django sample

pierrei:
Added a requirements file to show which packages are needed to be installed for the Django example to actually work.

To set up a virtual environment with these packages do the following:

```
virtualenv --no-site-packages virtualenv
. virtualenv/bin/activate
pip install -r requirements.txt
```

### DIFF
--- a/samples/django_sample/requirements.txt
+++ b/samples/django_sample/requirements.txt
@@ -1,0 +1,2 @@
+Django==1.5.12
+oauth2client==2.2.0
--- a/samples/django_sample/requirements.txt
+++ b/samples/django_sample/requirements.txt
@@ -1,2 +1,3 @@
-Django==1.5.12
+Django==1.4.22
 oauth2client==2.2.0
+google-api-python-client


pierrei:
I signed it!

nathanielmanistaatgoogle:
Thank you for the contribution! Please squash your commits?

wasabigeek:
Is there a plan to update the sample for later Django versions?

pierrei:
@nathanielmanistaatgoogle: No problem! How do I do that when I've already pushed my commits? Can't you squash when you merge?

pierrei:
@wasabigeek: No plan for me at least. I just wanted to help people to make it work in the first place :)

waprin:
@wasabigeek @pierrei I have new Samples here: https://github.com/google/oauth2client/pull/636 , feel free to take a look/comment . I think Django samples belong more there then here since the only Django specific stuff is the auth (oauth2client used to be part of this repo). So my inclination is just to delete this sample and redirect users there. What do you think?

And @pierrei are you actually using Django 1.4? Or just trying to be helpful? Because if you are then maybe I shouldn't delete them.

busunkim96:
Django samples have been removed.

PR #272, opened 2016-08-27
Make Django sample compatible with Django 1.1.0 and oauth2client 3.0.0

aavnsh:
Changes that will make the sample compatible with django 1.10 and oauth2client 3.0.0.
- request.REQUEST is deprecated in django 1.9 [Fields Removed in 1.9](https://docs.djangoproject.com/en/1.8/internals/deprecation/#deprecation-removed-in-1-9)
- oauth2client moved CredentialField from django_orm to django_util

### DIFF
--- a/samples/django_sample/plus/models.py
+++ b/samples/django_sample/plus/models.py
@@ -2,7 +2,7 @@ from django.contrib import admin
 from django.contrib.auth.models import User
 from django.db import models
 
-from oauth2client.contrib.django_orm import CredentialsField
+from oauth2client.contrib.django_util.models import CredentialsField
 
 
 class CredentialsModel(models.Model):
--- a/samples/django_sample/plus/models.py
+++ b/samples/django_sample/plus/models.py
@@ -1,11 +1,7 @@
-import pickle
-import base64
-
 from django.contrib import admin
 from django.contrib.auth.models import User
 from django.db import models
 
-from oauth2client.contrib.django_orm import FlowField
 from oauth2client.contrib.django_orm import CredentialsField
 
 
--- a/samples/django_sample/plus/views.py
+++ b/samples/django_sample/plus/views.py
@@ -11,7 +11,7 @@ from django_sample.plus.models import CredentialsModel
 from django_sample import settings
 from oauth2client.contrib import xsrfutil
 from oauth2client.client import flow_from_clientsecrets
-from oauth2client.contrib.django_orm import Storage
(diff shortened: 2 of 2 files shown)


nathanielmanistaatgoogle:
Looks reasonable to me but I'd like @waprin to also review.

waprin:
LGTM, though I think this is the minimum for 1.10 as there's other stuff that's deprecated (but still supported) like `MIDDLEWARE_CLASSES` and `TEMPLATE_DIRS` that should be replaced. So if we decide to keep these samples much longer should make sure it's all up to date with 1.10 recommendations.

I am leaving for 2 weeks but when I get back would like to come up with a more cohesive plan on organizing these samples and the ones I'm trying to put in oauth2client. I *think* my new samples are better than this roll your own stuff since you just use the decorators and let the library handle getting the callback view logic right, but hoping I can get more feedback from people using the existing samples.

aavnsh:
@waprin Agree on the additional work required - When I looked at oauth2client code and saw that its ahead on the MIDDLEWARE_CLASSES [changes](https://github.com/google/oauth2client/blob/master/oauth2client/contrib/django_util/__init__.py#L344); I added `MIDDLEWARE_CLASSES = MIDDLEWARE` as a workaround in settings.py to make sample work until a newer version of oauth2client is released.
Its good to hear that the samples are getting merged. 
cc @nathanielmanistaatgoogle

jayanth1991:
Needs changes related to `manage.py` as well to make it compatible with Django v1.6+ due to certain functions being deprecated:
`The functions setup_environ() and execute_manager() will be removed from django.core.management. This also means that the old (pre-1.4) style of manage.py file will no longer work.`

There are a lot of broken parts with this jump from Django 1.4 -> 1.10, its better to just create a new project and salvage some of the code that is API specific.

chainoy:
Hi

waprin:
@chainoy hi, are you wondering about getting this PR merged?

jonparrott:
@waprin can we just delete this sample?

waprin:
@jonparrott sure but can we sync up after next about a general plan for Django samples especially as they relate to oauth?

jonparrott:
@waprin yes, let's.

sduskis:
@theacodes, it seems this PR is way out of date.  Also, are we just going to delete the samples?

theacodes:
Bu sun deleted them.

PR #293, opened 2016-10-13
upgrade to unified uritemplate 3.0.0 [closes #57]

### LINKED ISSUE #57: Switch from "uritemplate" to "uritemplate.py"
There are at least 2 URI template packages on the PyPI: [uritemplate](https://pypi.python.org/pypi/uritemplate/0.6) and [uritemplate.py](https://pypi.python.org/pypi/uritemplate.py/0.3.0). Both of these packages use `uritemplate` as the package name to be imported, which causes a conflict when both are installed (see https://github.com/sigmavirus24/uritemplate/issues/14).

I propose that this project depend on  `uritemplate.py` because:

- `uritemplate.py` is API-compatible with `uritemplate`, but the reverse is not true. I have confirmed that this library works with uritemplate.py.
- Other projects depend on `uritemplate.py`. For example, we are currently working on a project that depends on both `github3.py` (GitHub API client) and `google-api-python-client`. Installing both results in an immediate `ImportError` due to the `uritemplate` conflict.

This  is a simple, low-risk change that would aid compatibility with other projects.

Thoughts?

simon-weber:
As of uritemplate 3.0.0 and uritemplate.py 3.0.2, the two packages are equivalent.
There were no changes to functionality; see
https://github.com/sigmavirus24/uritemplate/blob/master/HISTORY.rst.

This resolves the conflict described in https://github.com/google/google-api-python-client/issues/57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Python 3 in production.
 # Third Party Libraries and Dependencies
 The following libraries will be installed when you install the client library:
 * [httplib2](https://github.com/httplib2/httplib2)
-* [uri-templates](https://github.com/uri-templates/uritemplate-py)
+* [uritemplate](https://github.com/sigmavirus24/uritemplate)
 
 For development you will also need the following libraries:
 * [WebTest](http://pythonpaste.org/webtest/)
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ install_requires = [
     'httplib2>=0.8,<1',
     'oauth2client>=1.5.0,<4.0.0',
     'six>=1.6.1,<2',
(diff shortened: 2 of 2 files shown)


simon-weber:
I signed it!

jonparrott:
Thanks, @simon-weber!

simon-weber:
Sure thing. Any idea when this will be released? I'm eager to get rid of the special cases we had for resolving this 😁

jonparrott:
Whenever I get the time to cut another release. Maybe sometime next week?

simon-weber:
Sounds good; thanks.

PR #294, opened 2016-10-14
Release v1.5.4

jonparrott:


nathanielmanistaatgoogle:
Am I reading the test results correctly as indicating that we're not yet compatible with `oauth2client` 4.0.0?

jonparrott:
@nathanielmanistaatgoogle apparently not! :D I'll have to revisit this later.

nathanielmanistaatgoogle:
Understood; thanks.

jonparrott:
@nathanielmanistaatgoogle this is fixed now and should be good to merge. The only tricky bit is that file_cache depended on locked_file, which we removed. The cache stuff was already well-guarded against failure, but the tests were not.

PR #295, opened 2016-10-15
BatchError is unprintable using default constructor (#165)

### LINKED ISSUE #165: BatchError is unprintable using default constructor (one string)
This one should be pretty simple, I hope.

Here's the constructor signature: `def __init__(self, reason, resp=None, content=None):`, which doesn't require `resp` to be defined, and I can see it is not defined most of the time, for example, in googleapiclient/http.py.

Then, given the representation method:

```
def __repr__(self):
      return '<BatchError %s "%s">' % (self.resp.status, self.reason)
```

Which is also the string method:

```
__str__ = __repr__
```

This results in unprintable exceptions where `resp` is undefined, which is not very helpful when attempting to understand the error (e.g. #164).

ittus:
Fix #165

### DIFF
--- a/googleapiclient/errors.py
+++ b/googleapiclient/errors.py
@@ -126,6 +126,9 @@ class BatchError(HttpError):
     self.reason = reason
 
   def __repr__(self):
+    if getattr(self.resp, 'status', None) is None:
+      return '<BatchError "%s">' % (self.reason)
+    else:
       return '<BatchError %s "%s">' % (self.resp.status, self.reason)
 
   __str__ = __repr__
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -1053,7 +1053,9 @@ class TestBatch(unittest.TestCase):
     upload = MediaFileUpload(
         datafile('small.png'), chunksize=500, resumable=True)
     self.request1.resumable = upload
-    self.assertRaises(BatchError, batch.add, self.request1, request_id='1')
+    with self.assertRaises(BatchError) as batch_error:
+      batch.add(self.request1, request_id='1')
+    str(batch_error.exception)
 
   def test_execute_empty_batch_no_http(self):
     batch = BatchHttpRequest()


googleapiclient/errors.py
```diff
@@ -126,7 +126,10 @@ def __init__(self, reason, resp=None, content=None):
     self.reason = reason
 
   def __repr__(self):
+    if self.resp is not None and hasattr(self.resp, 'status'):
```
nathanielmanistaatgoogle: You should be aware that there is a movement gaining traction saying that [`hasattr` should never be used](http://doc.bazaar.canonical.com/developers/code-style.html#hasattr-and-getattr).

nathanielmanistaatgoogle: Why check the `None`-ness of `self.resp`? Why not just do the `hasattr` (or `getattr`) call?

nathanielmanistaatgoogle:
Also be sure that your commit log message conforms to [the guidelines](http://chris.beams.io/posts/git-commit/#seven-rules).

googleapiclient/errors.py
ittus: If self.resp is None, getattr(None, 'status') will raise Error. I have changed the code in new commit.

ittus: Thanks, I've changed to `getattr`

ittus:
Thank for your review. I've commit new changes.

googleapiclient/errors.py
nathanielmanistaatgoogle: `getattr(None, 'status')` will raise an exception, but what about `getattr(None, 'status', None)`?

ittus: Thank you. I've updated my code.

tests/test_http.py
```diff
@@ -1053,7 +1053,10 @@ def test_add_fail_for_resumable(self):
     upload = MediaFileUpload(
         datafile('small.png'), chunksize=500, resumable=True)
     self.request1.resumable = upload
-    self.assertRaises(BatchError, batch.add, self.request1, request_id='1')
+    with self.assertRaises(BatchError) as batch_error:
+      batch.add(self.request1, request_id='1')
+    self.assertEqual(str(batch_error.exception), '<BatchError "Media requests cannot be used in a batch request.">')
```
nathanielmanistaatgoogle: Drop this line? I don't think we care that the string value of the exception raised by the system under test is some exact value.

googleapiclient/errors.py
```diff
@@ -126,7 +126,10 @@ def __init__(self, reason, resp=None, content=None):
     self.reason = reason
 
   def __repr__(self):
+    if getattr(self.resp, 'status', None) is not None:
```
nathanielmanistaatgoogle: Drop the `not` and flip the branches.

ittus: I don't really get it. This code will only print `self.resp.status` if `self.resp.status` is not None.

nathanielmanistaatgoogle: I'm suggesting that you change
```
if <expression> is not None:
  return <A>
else:
  return <B>
```
to
```
if <expression> is None:
  return <B>
else:
  return <A>
```
. Doesn't that have (one syntax element's worth) less complexity with identical behavior?

tests/test_http.py
ittus: I'd like to test if `batch.add(self.request1, request_id='1')` raise `BatchError` here, it should not be `BatchError: <unprintable BatchError object>` instead of correct message.

googleapiclient/errors.py
ittus: Thanks. I got it.

tests/test_http.py
nathanielmanistaatgoogle: I absolutely agree that we should test that `batch.add(self.request1, request_id='1') raises `BatchError`. I'm suggesting that we don't want to write a test that fails if we later decide to word the error message slightly differently.

Tests form part of the definition of correctness of the system under test. I'm saying that we don't want to consider our system incorrect if the message associated with the exception that is raised is different than the message it currently raises.

ittus: Can I change to:
```
    with self.assertRaises(BatchError) as batch_error:
      batch.add(self.request1, request_id='1')
    try:
      print(str(batch_error.exception))
    except AttributeError:
      self.fail("BatchError repr raised AttributeError unexpectedly!")
```

ittus: Thanks, I've pushed new code.

nathanielmanistaatgoogle:
Change content looks good; please squash into one commit with a commit log message that conforms to [the guidelines](http://chris.beams.io/posts/git-commit/#seven-rules).

PR #298, opened 2016-10-21
Refresh all docs

saicheems:
Change the discovery query to fetch all discovery docs, not just the
preferred ones.


jonparrott:
Great, now we'll have 900 files every release instead of 200.

PR #312, opened 2016-11-28
Fix error in resumable upload handler

matthewayne:
The client library depends on the range header being in the response when the HTTP response code is 308, but the range header is not present in the response when the server hasn't received any bytes previously. This PR defaults to 0 in that case.


jonparrott:
@matthewayne it would be great if you could also update the test cases to cover this situation.

PR #316, opened 2016-12-14
Clarify supported versions of Python 3

chrish42:
Change the wording to explicitly mention that Python 3.3 and 3.4 are supported (instead of saying "3.3+"), as these are the versions that are tested in the tox.ini file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ See the [Developers Guide](https://developers.google.com/api-client-library/pyth
 # Python Version
 Python 2.6 or 2.7 are fully supported.
 
-Python 3.3+ is also now supported! However, this library has not yet been used
+Python 3.3 & 3.4 are also now supported! However, this library has not yet been used
 as thoroughly with Python 3, so we'd recommend testing before deploying with
 Python 3 in production.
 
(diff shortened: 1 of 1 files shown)


chrish42:
I signed the CLA.

README.md
```diff
@@ -18,7 +18,7 @@ See the [Developers Guide](https://developers.google.com/api-client-library/pyth
 # Python Version
 Python 2.6 or 2.7 are fully supported.
 
-Python 3.3+ is also now supported! However, this library has not yet been used
+Python 3.3 & 3.4 are also now supported! However, this library has not yet been used
```
jonparrott: Haha, can we update this entire paragraph while we're here?

```
Python 3.3 and 3.4 are supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions.
```

chrish42:
Haha. Sure! I can even update the tox.ini file to test on 3.5 too, if you want. :-) 

A bit more seriously, testing on 3.5 would be really nice, because it's much more interesting than 3.4 for both scientific computing (matrix multiply operator) and writing server software (async & await).

jonparrott:
Thanks, @chrish42

jonparrott:
> I can even update the tox.ini file to test on 3.5 too, if you want. :-)


@chrish42 be my guest. It's been on my list but I've been busy with `google-auth` and haven't had time to circle back to this library.

PR #320, opened 2016-12-26
Set default timeout for all the HTTP requests

i-maravic:
This will avoid having hanging connections when using Google API calls.

In case the user have set default socket timeout, that value is used instead of default.

### DIFF
--- a/describe.py
+++ b/describe.py
@@ -34,7 +34,7 @@ from googleapiclient.discovery import DISCOVERY_URI
 from googleapiclient.discovery import build
 from googleapiclient.discovery import build_from_document
 from googleapiclient.discovery import UnknownApiNameOrVersion
-import httplib2
+from googleapiclient.http import build_http
 import uritemplate
 
 CSS = """<style>
@@ -346,6 +346,7 @@ def document_api(name, version):
     print 'Warning: {} {} found but could not be built.'.format(name, version)
     return
 
+  http = build_http()
   response, content = http.request(
       uritemplate.expand(
           FLAGS.discovery_uri_template, {
@@ -366,7 +367,7 @@ def document_api_from_discovery_document(uri):
   Args:
     uri: string, URI of discovery document.
   """
-  http = httplib2.Http()
+  http = build_http()
   response, content = http.request(FLAGS.discovery_uri)
   discovery = json.loads(content)
 
@@ -384,7 +385,7 @@ if __name__ == '__main__':
   if FLAGS.discovery_uri:
     document_api_from_discovery_document(FLAGS.discovery_uri)
   else:
-    http = httplib2.Http()
+    http = build_http()
     resp, content = http.request(
         FLAGS.directory_uri,
         headers={'X-User-IP': '0.0.0.0'})
--- a/googleapiclient/_auth.py
+++ b/googleapiclient/_auth.py
@@ -14,8 +14,6 @@
 
 """Helpers for authentication using oauth2client or google-auth."""
 
-import httplib2
-
 try:
     import google.auth
     import google_auth_httplib2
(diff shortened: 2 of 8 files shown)


i-maravic:
I signed it!

i-maravic:
@fabriziodemaria

i-maravic:
@jonparrott Could you take a look at this PR please?

jonparrott:
5 minutes seems way too high. App Engine will terminate requests after 60 seconds.

i-maravic:
@jonparrott I wanted to be conservative.

I've changed it to 60s now.

jonparrott:
@i-maravic I guess my concern here is that I don't like the idea of using any library-specific timeout that the user doesn't explicitly set. It's surprising behavior. What prompted this change? Can I get some more context to better understand what problem this is trying to solve?

i-maravic:
@jonparrott We're using google-api-python-client when running periodic (hourly or daily) batch jobs for uploading to BigQuery and reading/writing to Google Cloud Storage. All the periodic jobs are being executed in Docker containers.

Recently we've experienced issues with those jobs hanging forever without making any progress. After doing some investigation we detected that the code making requests to Google APIs (aka. google-api-python-client) is hanging forever while waiting something to read from TCP socket. This was surprising behaviour for us.

After doing even more digging we discovered that Docker had some weird networking issues, which were causing this in the first place. We're not 100% sure how to reproduce Docker issue though, but restarting the Docker service mitigated a lot of problems we're experiencing.

I fell that library doing HTTP requests should set a sane default timeout for doing HTTP requests, because we could easily avoid issues like this in the future. I would argue that this is what most of the users would like to have any how.

I would also like to point out that you already have code in your lib, which is dealing with Timeout exceptions that might be thrown from the HTTP requests and do the retries accordingly.

I know that you're not using _requests_ library, but there is very nice [explanation](http://docs.python-requests.org/en/master/user/quickstart/#timeouts) why all production code using it should set timeouts for all the HTTP requests it's doing.

jonparrott:
@i-maravic I see, okay.

I need the CLA bot to verify your signature status before I can merge.

i-maravic:
I work for Spotify. Spotify should have signed the CLA.

jonparrott:
@i-maravic gotcha.

@lukesneeringer @landrito I'd appreciate if one of you could chime in and let me know your thoughts before I merge this.

lukesneeringer:
A 60 second default timeout seems reasonable to me, and it is easy enough to override. LGTM.

jonparrott:
@lukesneeringer thanks for chiming in!

@i-maravic okay, so it looks like this is going to happen, can I ask two things of you:

1. Is it possible to add tests or assertions to existing tests to cover this?
2. This timeout only applies to the http object used to fetch discovery documents. Do you want to apply the timeout to the http object used to make actual API requests as well?

i-maravic:
@jonparrott Sounds like reasonable asks.

Could you point me out where is the http object for actual API requests constructed?

jonparrott:
@i-maravic just below in `build_from_document`: https://github.com/google/google-api-python-client/pull/320/files#diff-bceb3b293cf7c789403d137d70ef8dd3R370

i-maravic:
@jonparrott Now the default timeout is applied to all discovery and API requests (auth or not)

Hopefully I didn't miss and httplib2.Http() instantiation in the lib

jonparrott:
@i-maravic this looks good, one last question:

Do you want to pass in `timeout` to `authorized_http` [here](https://github.com/google/google-api-python-client/pull/320/files#diff-bceb3b293cf7c789403d137d70ef8dd3R369)? Otherwise, the authorized http instance won't have the timeout.

i-maravic:
@jonparrott `authorized_http` gets timeout set [here](https://github.com/google/google-api-python-client/pull/320/files#diff-39e7a8494b350ee804d3775fae23de87R88)

jonparrott:
@i-maravic gotcha. Alright, this LGTM. Thanks!

PR #328, opened 2017-01-11
Adding import of google.auth.credentials to googleapiclient._auth module

wwlian:
Fixes https://github.com/google/google-api-python-client/issues/327


jonparrott:
Nice catch, thank you!

jonparrott:
This is released in 1.6.1

PR #330, opened 2017-01-13
Fix generation for methods with abnormal page token conventions

tcoffee-google:
Addresses https://github.com/googleapis/toolkit/issues/692.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -115,6 +115,7 @@ MEDIA_MIME_TYPE_PARAMETER_DEFAULT_VALUE = {
     'type': 'string',
     'required': False,
 }
+_PAGE_TOKEN_NAMES = ('pageToken', 'nextPageToken')
 
 # Parameters accepted by the stack, but not visible via discovery.
 # TODO(dhermes): Remove 'userip' in 'v2'.
@@ -720,7 +721,7 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
       if name not in kwargs:
         # temporary workaround for non-paging methods incorrectly requiring
         # page token parameter (cf. drive.changes.watch vs. drive.changes.list)
-        if name not in pageTokenNames or pageTokenName(
+        if name not in _PAGE_TOKEN_NAMES or findPageTokenName(
             methodProperties(methodDesc, schema, 'response')):
           raise TypeError('Missing required parameter "%s"' % name)
 
@@ -925,16 +926,18 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
   return (methodName, method)
 
 
-def createNextMethod(methodName, pageToken, nextPageToken,
-                     isPageTokenParameter):
+def createNextMethod(methodName,
+                     pageTokenName='pageToken',
+                     nextPageTokenName='nextPageToken',
+                     isPageTokenParameter=True):
   """Creates any _next methods for attaching to a Resource.
 
   The _next methods allow for easy iteration through list() responses.
 
   Args:
     methodName: string, name of the method to use.
-    pageToken: string, name of request page token field.
-    nextPageToken: string, name of response page token field.
+    pageTokenName: string, name of request page token field.
+    nextPageTokenName: string, name of response page token field.
     isPageTokenParameter: Boolean, True if request page token is a query
         parameter, False if request page token is a field of the request body.
   """
@@ -954,28 +957,24 @@ Returns:
     # Retrieve nextPageToken from previous_response
     # Use as pageToken in previous_request to create new request.
 
-    if nextPageToken not in previous_response or not previous_response[nextPageToken]:
+    if (nextPageTokenName not in previous_response or
+        not previous_response[nextPageTokenName]):
       return None
 
     request = copy.copy(previous_request)
 
-    token = previous_response[nextPageToken]
+    nextPageToken = previous_response[nextPageTokenName]
 
     if isPageTokenParameter:
         # Replace pageToken value in URI
-        parsed = list(urlparse(request.uri))
-        q = parse_qsl(parsed[4])
-        newq = [(key, value) for (key, value) in q if key != pageToken]
-        newq.append((pageToken, token))
-        parsed[4] = urlencode(newq)
-        uri = urlunparse(parsed)
-        request.uri = uri
-        logger.info('Next page request URL: %s %s' % (methodName, uri))
+        request.uri = _add_query_parameter(
+            request.uri, pageTokenName, nextPageToken)
+        logger.info('Next page request URL: %s %s' % (methodName, request.uri))
     else:
         # Replace pageToken value in request body
         model = self._model
         body = model.deserialize(request.body)
-        body[pageToken] = token
+        body[pageTokenName] = nextPageToken
         request.body = model.serialize(body)
         logger.info('Next page request body: %s %s' % (methodName, body))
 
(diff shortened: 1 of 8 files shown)


tcoffee-google:
@jonparrott @vchudnov-g @saicheems

tcoffee-google:
Note this fix should be submitted in concert with https://github.com/googleapis/toolkit/pull/880.

googleapiclient/discovery.py
```diff
@@ -921,13 +925,18 @@ def method(self, **kwargs):
...
+def createNextMethod(methodName, pageToken, nextPageToken,
+                     isPageTokenParameter):
   """Creates any _next methods for attaching to a Resource.
 
   The _next methods allow for easy iteration through list() responses.
 
   Args:
     methodName: string, name of the method to use.
+    pageToken: string, name of request page token field.
+    nextPageToken: string, name of response page token field.
```
jonparrott: Would `nextPageTokenField` or `nextPageTokenFieldName` be a better fit?

(Also I really hate this lowerCamelCase nonsense, but I know you're just fitting in with existing style)

Should this have a default value?

googleapiclient/discovery.py
```diff
@@ -945,24 +954,30 @@ def methodNext(self, previous_request, previous_response):
...
-    uri = urlunparse(parsed)
-
-    request.uri = uri
-
-    logger.info('URL being requested: %s %s' % (methodName,uri))
+    token = previous_response[nextPageToken]
+
+    if isPageTokenParameter:
+        # Replace pageToken value in URI
+        parsed = list(urlparse(request.uri))
```
jonparrott: I think there's an existing helper method for updating a query string.

googleapiclient/discovery.py
```diff
@@ -1110,19 +1125,36 @@ def methodResource(self):
                                method.__get__(self, self.__class__))
 
   def _add_next_methods(self, resourceDesc, schema):
-    # Add _next() methods
-    # Look for response bodies in schema that contain nextPageToken, and methods
-    # that take a pageToken parameter.
+    # Add _next() methods iff one of the names 'pageToken' or 'nextPageToken'
```
jonparrott: I know `iff` is popular with Googlers, but can we use `if and only if`.

googleapiclient/discovery.py
```diff
@@ -1110,19 +1125,36 @@ def methodResource(self):
...
-        if 'response' in methodDesc:
-          responseSchema = methodDesc['response']
-          if '$ref' in responseSchema:
-            responseSchema = schema.get(responseSchema['$ref'])
-          hasNextPageToken = 'nextPageToken' in responseSchema.get('properties',
-                                                                   {})
-          hasPageToken = 'pageToken' in methodDesc.get('parameters', {})
-          if hasNextPageToken and hasPageToken:
-            fixedMethodName, method = createNextMethod(methodName + '_next')
+        nextPageToken = pageTokenName(
```
jonparrott: Since you're here, why not fix this code to not be quite so nested? Early return clauses would make this much more readable.

googleapiclient/discovery.py
```diff
@@ -1110,19 +1125,36 @@ def methodResource(self):
...
+                methodProperties(methodDesc, schema, 'request'))
+          if pageToken:
+            fixedMethodName, method = createNextMethod(
+                methodName + '_next', pageToken, nextPageToken,
+                isPageTokenParameter)
             self._set_dynamic_attr(fixedMethodName,
                                    method.__get__(self, self.__class__))
+
+
+pageTokenNames = ('pageToken', 'nextPageToken')
```
jonparrott: Module-level constants go at the top and should be `UPPER_SNAKE_CASE`. Also, there's no need for this to be public so please do `_PAGE_TOKEN_FIELD_NAMES`.

googleapiclient/discovery.py
```diff
@@ -1110,19 +1125,36 @@ def methodResource(self):
...
+
+def pageTokenName(fields):
```
jonparrott: Docstrings for these two functions, please.

googleapiclient/discovery.py
tcoffee-google: Used `...TokenName`.

Yep.

Done.

googleapiclient/discovery.py
tcoffee-google: Yep --- this was old code; seemed oddly low-level here, but I didn't search the oauth2 library.

googleapiclient/discovery.py
tcoffee-google: Done.

googleapiclient/discovery.py
tcoffee-google: Not sure I agree in cases neither exceptional nor degenerate, but fine.

googleapiclient/discovery.py
tcoffee-google: Done.

googleapiclient/discovery.py
tcoffee-google: Done.

googleapiclient/discovery.py
```diff
@@ -1110,19 +1124,59 @@ def methodResource(self):
...
+      if not pageTokenName:
+        continue
+      fixedMethodName, method = createNextMethod(
+          methodName + '_next', pageTokenName, nextPageTokenName,
+          isPageTokenParameter)
+      self._set_dynamic_attr(fixedMethodName,
+                             method.__get__(self, self.__class__))
+
+
+def findPageTokenName(fields):
```
jonparrott: If this is private, please prefix it with `_`.

googleapiclient/discovery.py
```diff
@@ -1110,19 +1124,59 @@ def methodResource(self):
...
+    fields: iterable, names of fields.
+
+  Returns:
+    First name that is either 'pageToken' or 'nextPageToken' if one exists,
+    otherwise None.
+  """
+  return next((tokenName for tokenName in _PAGE_TOKEN_NAMES
+              if tokenName in fields), None)
+
+def methodProperties(methodDesc, schema, name):
```
jonparrott: I think this is also intended to be private.

googleapiclient/model.py
```diff
@@ -348,7 +348,7 @@ def no_content_response(self):
 def makepatch(original, modified):
   """Create a patch object.
 
-  Some methods support PATCH, an efficient way to send updates to a resource.
+  Some methods support PATCH, an efficient way to send updates to a presource.
```
jonparrott: Typo?

googleapiclient/schema.py
```diff
@@ -161,13 +161,13 @@ def prettyPrintSchema(self, schema):
     # Return with trailing comma and newline removed.
     return self._prettyPrintSchema(schema, dent=1)[:-2]
 
-  def get(self, name):
+  def get(self, name, default=None):
```
jonparrott: Update the docstring if you add a new parameter, please.

googleapiclient/discovery.py
tcoffee-google: Done.

googleapiclient/discovery.py
tcoffee-google: Done.

googleapiclient/model.py
tcoffee-google: I assume.

googleapiclient/schema.py
tcoffee-google: Done.

googleapiclient/model.py
jonparrott: Fix it? (Or if you did, i'm not seeing the updated file)

jonparrott:
@lukesneeringer @landrito I'd love to have one of you take a look.

Large changes like this on this ancient codebase make me uneasy.

googleapiclient/model.py
tcoffee-google: Oh, sorry --- I had fixed this earlier and it somehow got reverted.

googleapiclient/discovery.py
```diff
@@ -718,7 +719,11 @@ def method(self, **kwargs):
 
     for name in parameters.required_params:
       if name not in kwargs:
-        raise TypeError('Missing required parameter "%s"' % name)
+        # temporary workaround for non-paging methods incorrectly requiring
+        # page token parameter (cf. drive.changes.watch vs. drive.changes.list)
+        if name not in _PAGE_TOKEN_NAMES or _findPageTokenName(
```
lukesneeringer: What is the ultimate goal here? This method needs to _not_ require a `pageToken` or `nextPageToken`?

I am really struggling to follow this `if` block:

  * If `pageToken` is set as a required parameter and not provided, then the first disjunct will fail, and then `_findPageTokenName` will return a string, which is truthy (even if the string is `nextPageToken`), and we get an error. That may or may not be correct behavior (and sure does not seem like it is correct behavior on purpose).
  * If `foo` is a required parameter, then it is not in `_PAGE_TOKEN_NAMES`, and the first disjunct is true, and you get the exception. Okay, fine.

googleapiclient/discovery.py
```diff
@@ -945,24 +957,26 @@ def methodNext(self, previous_request, previous_response):
     # Retrieve nextPageToken from previous_response
     # Use as pageToken in previous_request to create new request.
 
-    if 'nextPageToken' not in previous_response or not previous_response['nextPageToken']:
+    if (nextPageTokenName not in previous_response or
+        not previous_response[nextPageTokenName]):
```
lukesneeringer:
```python
if not previous_response.get(nextPageTokenName, None):
```

googleapiclient/discovery.py
```diff
@@ -1110,19 +1124,59 @@ def methodResource(self):
...
+  """Search field names for one like a page token.
+
+  Args:
+    fields: iterable, names of fields.
+
+  Returns:
+    First name that is either 'pageToken' or 'nextPageToken' if one exists,
+    otherwise None.
+  """
+  return next((tokenName for tokenName in _PAGE_TOKEN_NAMES
```
lukesneeringer: Is there a particular reason why this is implemented as such? Does order of fields matter? I can see weird bugs cropping up if this is sent a partially-iterated iterable...

googleapiclient/discovery.py
tcoffee-google: This is a workaround for a bug that currently only affects the drive.changes.watch method. Its discovery doc says it requires a page token, but it's not a paging method at all. So the 'missing' error is valid if the parameter name is not a page token or if the method is actually paging (which at this point is equivalent to it returning a page token in its response). Would you prefer 

```python
if not (name in _PAGE_TOKEN_NAMES and not _findPageTokenName(...))
```
?

googleapiclient/discovery.py
tcoffee-google: While equivalent, I think the original makes it clearer that two distinct meaningful cases are covered here.

googleapiclient/discovery.py
tcoffee-google: Do you mean a partially-iterated iterator? I suppose I should say "container" rather than "iterable".

It's implemented this way because the actual calling cases make `fields` a dict.

vchudnov-g:
ping

googleapiclient/discovery.py
jonparrott: I'd prefer @lukesneeringer's suggestion.

tcoffee-google: Done.

tcoffee-google:
Any further comments on this?

lukesneeringer:
No further comments. Thanks!

tcoffee-google:
@jonparrott @lukesneeringer can you please approve changes to enable submission? Thanks!

Issue #342, opened 2017-01-31
Allow user to set user-agent for api queries


Drino:
Currently there is no way to set query user-agent when calling an API method. I want to add a parameter to re-write this field.

jonparrott:
@Drino can we get some context as to why this is needed? You can already set the user-agent at the http level and we have a default user agent set.

Drino:
You're right. It's not needed.

PR #344, opened 2017-01-31
Allow setting user-agent header for api methods.

Drino:
Allow setting user-agent header for api methods.

Now if one need to call an API method with his own user agent, he may
just use user_agent keyword argument for method call.

Resolves: #342

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -708,7 +708,7 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
 
   parameters = ResourceMethodParameters(methodDesc)
 
-  def method(self, **kwargs):
+  def method(self, user_agent=None, **kwargs):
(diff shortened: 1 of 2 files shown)


googleapiclient/discovery.py
```diff
@@ -708,7 +708,7 @@ def createMethod(methodName, methodDesc, rootDesc, schema):
 
   parameters = ResourceMethodParameters(methodDesc)
 
-  def method(self, **kwargs):
+  def method(self, user_agent=None, **kwargs):
```
jonparrott: My biggest worry with this change is that it's possible an API method will have a `user_agent` field and this will prevent that from ever being possible to set.

Drino:
I've found that my problems can be solved another way. Sorry for making a pull request.

jonparrott:
No worries. Always happy to take a look.

PR #347, opened 2017-02-07
Don't invoke application default credentials logic when developerKey is provided

### LINKED ISSUE #346: v1.6.1 HttpError 403 "Insufficient Permission"
I happened to try updating google-api-python-client from 1.5.5 to 1.6.1. It causes my YouTube Data API v3 code to always fail with a HttpError. An example to reproduce is:

    DEVELOPER_KEY = "<Insert key>"
    youtube = build("youtube", "v3", developerKey=DEVELOPER_KEY)
    youtube.videos().list(id="xGG56MWgbOA", part="id").execute()

This returns a successful response on 1.5.5, but raises a HttpError on 1.6.1:

    HttpError: <HttpError 403 when requesting https://www.googleapis.com/youtube/v3/videos?alt=json&part=id&id=xGG56MWgbOA&key=<Insert key> returned "Insufficient Permission">

Visiting the specified link in a browser returns a valid response. Is there some usage change I am not aware of?

jonparrott:
Resolves #346

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -355,8 +355,9 @@ def build_from_document(
     scopes = list(
       service.get('auth', {}).get('oauth2', {}).get('scopes', {}).keys())
 
-    # If so, then the we need to setup authentication.
-    if scopes:
+    # If so, then the we need to setup authentication if no developerKey is
+    # specified.
+    if scopes and not developerKey:
       # If the user didn't pass in credentials, attempt to acquire application
       # default credentials.
       if credentials is None:
--- a/tests/test_discovery.py
+++ b/tests/test_discovery.py
@@ -437,6 +437,16 @@ class DiscoveryFromDocument(unittest.TestCase):
       discovery, base="https://www.googleapis.com/", http=http)
     self.assertEquals(plus._http, http)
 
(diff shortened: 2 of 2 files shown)


googleapiclient/discovery.py
```diff
@@ -355,8 +355,9 @@ def build_from_document(
     scopes = list(
       service.get('auth', {}).get('oauth2', {}).get('scopes', {}).keys())
 
-    # If so, then the we need to setup authentication.
-    if scopes:
+    # If so, then the we need to setup authentication if no developerKey is
+    # specified.
+    if scopes and not developerKey:
```
lukesneeringer: If a `developerKey` is specified, does this make the value of `scopes` no longer relevant?
If so, is there a case for raising an exception if both are specified (and scopes is non-default)?

lukesneeringer:
Approved with the one question.

googleapiclient/discovery.py
jonparrott: scopes can't be passed in, it's gather from the discovery doc. Scopes indicate that the service requires some sort of authentication, but if the user provides a developerKey then authentication is handled by that.

Issue #350, opened 2017-02-10
`TypeError` when trying to issue `BatchHttpRequest` when service was built using application default credentials


mgilson:
This issue is very similar to #211.  My team has an app-engine application that uses batch requests in the Gmail API.  After an upgrade of this client library, we started seeing failures:

```
...
  File "/base/data/home/apps/s~app-id/modname:version/path/to/my/code.py", line 241, in _user_method
    return batch_request.execute(http=self._http)
  File "/base/data/home/apps/s~app-id/modname:version/third_party/oauth2client/_helpers.py", line 133, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/base/data/home/apps/s~app-id/modname:version/third_party/googleapiclient/http.py", line 1417, in execute
    self._execute(http, self._order, self._requests)
  File "/base/data/home/apps/s~app-id/modname:version/third_party/googleapiclient/http.py", line 1333, in _execute
    body = self._serialize_request(request)
  File "/base/data/home/apps/s~app-id/modname:version/third_party/googleapiclient/http.py", line 1204, in _serialize_request
    request.http.request.credentials.apply(headers)
  File "/base/data/home/apps/s~app-id/modname:version/third_party/oauth2client/client.py", line 558, in apply
    headers['Authorization'] = 'Bearer ' + self.access_token
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

Our usage of this client library looks like the following:

```python
from googleapiclient.discovery import build
gmail_service = build('gmail', 'v1')  # Uses application credentials

...

authorized_http = user_credentials.authorize(httplib2.Http())  # Uses end user credentials

batch_request = gmail_service.new_batch_http_request()
batch_request.add(
    gmail_service.users().threads().get(id=thread_id, userId='me'),
    callback=callback,
    request_id=thread_id)

batch_request.execute(http=authorized_http)
```

The `gmail_service` is cached at the application level to avoid needing to do an API call for each request (which is why it ends up with the application credentials).  What I believe is happening is that

```py
gmail_service.users().threads().get(id=thread_id, userId='me')
```

ends up creating an `HttpRequest` that uses the same `http` that was used to construct `gmail_service` (i.e. the application credentials) and I believe that the application credentials have no `access_token`.  This results in the `TypeError` seen above.  The fix in [#232](https://github.com/google/google-api-python-client/pull/232/files) won't help (the `http` that is getting passed to `batch_request.execute` is valid).

[...]

PR #351, opened 2017-02-14
App default creds

mgilson:
Only authenticate individual requests in a batch if the credentials are non-`None`.  Otherwise, assume that an appropriate authentication is being provided at a higher level (e.g. for the entire batch)

This should address #350

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1202,7 +1202,17 @@ class BatchHttpRequest(object):
 
     if request.http is not None and hasattr(request.http.request,
         'credentials'):
-      request.http.request.credentials.apply(headers)
+      credentials = request.http.request.credentials
+      if (hasattr(credentials, 'access_token') and
+          credentials.access_token is None):
+        # It is possible to have credentials and still not have an access token.
+        # (e.g. with application default credentials on GAE).  If that is the
+        # case, do not serialize the authentication headers with this request.
+        # Instead, assume that the batch has an authorization header which will
+        # trickle down to all of the individual requests.
+        pass
+      else:
+        credentials.apply(headers)
 
     # MIMENonMultipart adds its own Content-Type header.
     if 'content-type' in headers:
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -88,7 +88,7 @@ def _should_retry_response(resp_status, content):
 
   Args:
     resp_status: The response status received.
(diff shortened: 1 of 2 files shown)


googleapiclient/http.py
```diff
@@ -1202,7 +1202,17 @@ def _serialize_request(self, request):
 
     if request.http is not None and hasattr(request.http.request,
         'credentials'):
-      request.http.request.credentials.apply(headers)
+      credentials = request.http.request.credentials
```
mgilson: I've played around with a couple different implementations here and the tests have led me to recognize a few other corner cases and/or places where this isn't ideal.  Specifically, 

* `Credentials` doesn't have an `access_token` property.
* `OAuth2Credentials` does have an `access_token` property.
* `AssertionCredentials` (which eventually subclass `OAuth2Credentials` _do_ have an `access_token` property, but it gets set to `None` in `__init__` which leaves `AssertionCredentials` with a non-working `apply` method
  - `apply` looks to be inherited from `OAuth2Credentials` which assumes that `access_token` is a string

I'm not super familiar with the architecture of this library and the architecture of the `oauth2client` library and how much their requirements are supposed to interoperate.  Having looked at this more, it feels like the bug might actually be in `oauth2client.client.AssertionCredentials.apply`.  If that method were to become a no-op (at least if `self.access_token is None`), then the problems _here_ would go away.  I'm not sure if that's the right thing to do _there_ though.

Maybe this isn't the right place to raise that issue (but it looks like a lot of the committers to both libraries are the same) -- and I suppose the fix here would still need to be in place to handle interoperability with past versions of that library.  Anyway, any advice you can give on this front is welcome :-)

jonparrott:
@mgilson thanks for doing this- heads up I'm pretty busy the next couple of days but I will try to get to this by friday.

mgilson:
Thanks for letting me know when to be looking into this again.  I'm in no hurry -- We've already coded the workaround that I mentioned in the issue on our end so there's nothing mission critical for us.  In other words, take your time.  :-)

jonparrott:
@mgilson sorry I got super backlogged on this. I think it might not be as easy as we think - this library has to support both oauth2client and google-auth. We might have to add some special logic to handle those. Let me find some time to look into this in more depth to see what's up.

jonparrott:
Closing in favor of #376.

PR #352, opened 2017-02-18
Python 3.3 is no longer supported

hugovk:
For example, see its removal in https://github.com/google/google-api-python-client/pull/341.

TODO:

* [ ] These [docs](https://developers.google.com/api-client-library/python/apis/customsearch/v1) also need updating: "Python 2.6, 2.7, 3.3, or 3.4". Are they open source too?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,8 +25,8 @@ if sys.version_info < (2, 7):
   print('google-api-python-client requires python version >= 2.7.',
         file=sys.stderr)
   sys.exit(1)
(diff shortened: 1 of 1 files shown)


hugovk:
@googlebot I signed it!

jonparrott:
@hugovk I'd prefer us not to make the change in setup.py. I believe this can still be installed and used on 3.3, we're just not doing automated tests or supporting 3.3. Basically, install and use at your own risk on 3.3.

PR #376, opened 2017-03-30
Fix credentials usage in BatchHTTPRequest

### LINKED ISSUE #375: BatchHttpRequest and google-auth
Hi,

I'm using this client library with google-auth to make calls to several Google APIs (Drive mainly), it works fine however it seems that when using a Batch request it doesn't use the credentials of each request in the batch, only the credentials from the first one.

Which means that when I impersonate two different users in two different requests in the same batch, both requests are made impersonating the first user.

For example here's how I proceed to list the files of two users:

```
creds_one = google.oauth2.service_account.Credentials.from_service_account_file(path_to_json, scopes=scopes, subject='user1@xxx.com')
creds_two = google.oauth2.service_account.Credentials.from_service_account_file(path_to_json, scopes=scopes, subject='user2@xxx.com')

drive_service_one = build('drive', 'v3', credentials=creds_one)
drive_service_two = build('drive', 'v3', credentials=creds_two)

_batch = BatchHttpRequest()
_batch.add(request=drive_service_one.files().list(), request_id='u1')
_batch.add(request=drive_service_two.files().list(), request_id='u2')

_batch.execute()
```

What I observe in this case is that both requests return the list of files from the first user.
Am I missing something here ?

Thanks

jonparrott:
Resolves #375 and #350.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1138,7 +1138,7 @@ class BatchHttpRequest(object):
 
     if creds is not None:
       if id(creds) not in self._refreshed_credentials:
-        _auth.refresh_credentials(creds, http)
+        _auth.refresh_credentials(creds)
         self._refreshed_credentials[id(creds)] = 1
 
     # Only apply the credentials if we are using the http object passed in,
@@ -1208,8 +1208,7 @@ class BatchHttpRequest(object):
     if request.http is not None:
       credentials = _auth.get_credentials_from_http(request.http)
       if credentials is not None:
-        if _auth.has_access_token(credentials):
-          _auth.apply_credentials(credentials, headers)
+        _auth.apply_credentials(credentials, headers)
 
     # MIMENonMultipart adds its own Content-Type header.
     if 'content-type' in headers:
@@ -1418,9 +1417,9 @@ class BatchHttpRequest(object):
     # refreshed with an initial access_token.
     creds = _auth.get_credentials_from_http(http)
     if creds is not None:
-      if not _auth.has_access_token(creds):
+      if not _auth.is_valid(creds):
         LOGGER.info('Attempting refresh to obtain initial access_token')
-        _auth.refresh_credentials(creds, http)
+        _auth.refresh_credentials(creds)
 
     self._execute(http, self._order, self._requests)
 
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -62,6 +62,7 @@ try:
 except ImportError:
   from oauth2client import _helpers as util
 
+from googleapiclient import _auth
 from googleapiclient import mimeparse
 from googleapiclient.errors import BatchError
 from googleapiclient.errors import HttpError
@@ -1126,21 +1127,25 @@ class BatchHttpRequest(object):
     # If there is no http per the request then refresh the http passed in
     # via execute()
     creds = None
-    if request.http is not None and hasattr(request.http.request,
-        'credentials'):
-      creds = request.http.request.credentials
-    elif http is not None and hasattr(http.request, 'credentials'):
-      creds = http.request.credentials
+    request_credentials = False
+
+    if request.http is not None:
+      creds = _auth.get_credentials_from_http(request.http)
+      request_credentials = True
+
+    if creds is None and http is not None:
+      creds = _auth.get_credentials_from_http(http)
+
     if creds is not None:
       if id(creds) not in self._refreshed_credentials:
-        creds.refresh(http)
+        _auth.refresh_credentials(creds, http)
         self._refreshed_credentials[id(creds)] = 1
 
     # Only apply the credentials if we are using the http object passed in,
     # otherwise apply() will get called during _serialize_request().
-    if request.http is None or not hasattr(request.http.request,
-        'credentials'):
-      creds.apply(request.headers)
+    if request.http is None or not request_credentials:
+      _auth.apply_credentials(creds, request.headers)
+
 
   def _id_to_header(self, id_):
     """Convert an id to a Content-ID header value.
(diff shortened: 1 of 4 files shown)


jonparrott:
Please don't merge until I verify this will fix #375 and have a chance to write tests.

sbrieuc:
@jonparrott Looks like there is a circular import issue

```
  File "/Users/brieuc/development/source/pip-libs/apiclient/__init__.py", line 19, in <module>
    from googleapiclient import discovery
  File "/Users/brieuc/development/source/pip-libs/googleapiclient/discovery.py", line 56, in <module>
    from googleapiclient import _auth
  File "/Users/brieuc/development/source/pip-libs/googleapiclient/_auth.py", line 32, in <module>
    from googleapiclient.http import build_http
  File "/Users/brieuc/development/source/pip-libs/googleapiclient/http.py", line 65, in <module>
    from googleapiclient import _auth
ImportError: cannot import name _auth
```
Now going beyond that, the _serialize_request method should also be updated with the _auth.get_credentials_from_http method, which leads to issue #350.

So I think we should change the _serialize_request from:
```python
        'credentials'):
    request.http.request.credentials.apply(headers)
```
to
```python
if request.http is not None:
    creds = _auth.get_credentials_from_http(request.http)
    if creds is not None:
        if _auth.has_access_token(creds):
            creds.apply(headers)
```

jonparrott:
@sbrieuc updated, can you give it another go? 
@mgilson it'd be great if you could try it out as well, as this should now resolve #350 as well.

googleapiclient/http.py
```diff
@@ -1200,9 +1205,10 @@ def _serialize_request(self, request):
     msg = MIMENonMultipart(major, minor)
     headers = request.headers.copy()
 
-    if request.http is not None and hasattr(request.http.request,
-        'credentials'):
-      request.http.request.credentials.apply(headers)
+    if request.http is not None:
+      credentials = _auth.get_credentials_from_http(request.http)
```
mgilson: There's a slight difference here compared to my PR.  It's been quite a while since I've looked at this, but the problem with GAE app-default credentials is that their `"access_token"` property is `None`.  This leads to problems when `credentials.apply(headers)` is called because the [`credentials.apply`](https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L556) does a string concatenation without checking anything.

The actual proof that the credentials don't have an `access_token` is hard to trace through the code, but I'll try.

The app-default credentials are of type [`AppAssertionCredentials`](https://github.com/google/oauth2client/blob/9f0618d53481a3afe2d0cf232131ba21dcf583de/oauth2client/contrib/appengine.py#L119) which inherits from [`AssertionCredentials`](https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L1444) which inherits from [`GoogleCredentials`](https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L1048).  And that finally gets us to inheriting from `OAuth2Credentials` -- Which is the first concrete subclass of `Credentials` in this inheritance tree.  `GoogleCredentials` is documented to expect a string access token (first parameter in `__init__`), but if you look at the code of  `AppAssertionCredentials`, it [unconditionally passes `None`](https://github.com/google/oauth2client/blob/master/oauth2client/client.py#L1473).

In order to get this to actually work on GAE, you need to put something in your `_auth` wrapper to make sure that `apply_credentials` becomes robust to the case where the credentials have a `None` `access_token` value.  In my initial PR, I checked that the `credentials.access_token is not None` before calling `credentials.apply(headers)`.  In principle you could still do that in your auth wrapper after checking that the credentials are coming from the `oauth2client` library and not the `google_auth` library ...

mgilson: I guess that's equivalent to checking `_auth.has_access_token` before calling `_auth.apply_credentials`.  It looks like you're already doing that on line 1420, maybe do the same thing here?

jonparrott: @mgilson Updated, take a look?

mgilson: @jonparrott -- Yep, that looks good.

sbrieuc:
@jonparrott Just tried this, everything's good !

sbrieuc:
@jonparrott Actually I'm seeing something that might be a bit off from a couple of tests I just performed. In the execute() method, in the case where we have no initial token, we manually refresh the credentials providing the http instance (either the one given to the execute method or the one from the first request in the batch):

```python
# Special case for OAuth2Credentials-style objects which have not yet been
# refreshed with an initial access_token.
creds = _auth.get_credentials_from_http(http)
if creds is not None:
  if not _auth.has_access_token(creds):
    LOGGER.info('Attempting refresh to obtain initial access_token')
    _auth.refresh_credentials(creds, http)

self._execute(http, self._order, self._requests)
```
Now in _auth.refresh_credentials method, in the case of google-auth credentials we do this:

```python
request = google_auth_httplib2.Request(http)
return credentials.refresh(request)
```
We're building a Request object using the provided http instance, which in this case will be an AuthorizedHttp object. That means it will use the AuthorizedHttp.request method to perform the request that will get an access token. However, this request method also performs a token refresh if needed. So what I observed is that when we have no initial token for google-auth credentials, the token gets refreshed twice: once because we manually called credentials.refresh and once because the request method used to refresh this token also performs a token refresh.

So my suggestion would be to use a new httplib2.Http() object for the request.

```python
request = google_auth_httplib2.Request(httplib2.Http())
return credentials.refresh(request)
```
This way the credentials are refreshed only once.

What do you think ?

jonparrott:
@sbrieuc I think that's pretty good idea. Updating now.

jonparrott:
I think this is now in a good state if you want to review, @lukesneeringer.

@sbrieuc I'd love if you can give this another test.

sbrieuc:
@jonparrott That seems good to me. All my tests are ok

jonparrott:
Great, @lukesneeringer PTAL?

jonparrott:
@lukesneeringer another poke here - this should be ready to merge.

jonparrott:
I'm going to take @lukesneeringer's silence as tacit approval. (I am happy to address concerns in a follow-up PR)

googleapiclient/http.py
```diff
@@ -1208,8 +1208,7 @@ def _serialize_request(self, request):
     if request.http is not None:
       credentials = _auth.get_credentials_from_http(request.http)
       if credentials is not None:
-        if _auth.has_access_token(credentials):
```
snarfed: damn, i should have tested my downstream usage of this fix before asking for #421. this `if` that @mgilson asked for in https://github.com/google/google-api-python-client/pull/376/commits/282fc889956e42de8ec3e26aecbe39509f027333#r109217447 was necessary to fix #350, at least in my case, and losing it here regresses and results in #350 happening for me again.

i've added it back in #434. apologies for not testing with this patch earlier!

PR #401, opened 2017-06-05
Clean and regen docs

saicheems:


saicheems:
Note, I cleaned out the docs directory before regenerating, meaning some old APIs which are no longer discoverable are not included (ex: dfareporting:v2.5).

Don't know if this is the recommended practice here, please let me know if not.

Issue #15, opened 2014-06-21
Bug in http.MediaIoBaseDownload


intellisense:
Problem is [here](https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L558). If `content-range` is not found in response. This will leads to infinite loop because `self._total_size` will always be `None` and `self._progress == self._total_size` will always results in `False` hence `self._done` will never be `True`. Shouldn't it raise an exception if `content-range` header is not found?

musalbas:
This is still an issue, and a serious one because it makes it impossible to download objects from Google Cloud Storage buckets using `MediaIoBaseDownload`.

Here's another way of getting objects (this does it in one go and stores it in memory, so may not be suitable for large files):
<pre>request = client.objects().get_media(bucket="blab", object="blab")
response = request.execute()
print response</pre>

markz0r:
This also applies to downloading via the Drive API, specifically mimeType Application/vnd.google-apps.*
Thanks @musalbas your advice solved my problem.

jonparrott:
I'm happy to accept/review any PRs to fix this issue.

Capstan:
We regularly test using `MediaIoBaseDownload` against Google Cloud Storage buckets, so I'm curious what your actual call pattern is to see what you might be doing differently. /cc @thobrla

markz0r:
@jonparrott ok thanks will take a look

@Capstan This code example fails as total_size is never defined, that status is always False. This only occurs for the export_media function which does not provide a 'size' field as the get_media function does. 
```
file_id = '1ZdR3L3qP4Bkq8noWLJHSr_iBau0DNT4Kli4SxNc2YEo'
request = drive_service.files().export_media(fileId=file_id,
                                             mimeType='application/pdf')
fh = io.BytesIO()
downloader = MediaIoBaseDownload(fh, request)
done = False
while done is False:
    status, done = downloader.next_chunk()
    print "Download %d%%." % int(status.progress() * 100)
```
source: [https://developers.google.com/drive/v3/web/manage-downloads#downloading_google_documents]

wor-k:
having the same issue as @markz0r 
Do we have any workaround to check completed exporting file?

eupharis:
I too believe I am running into this issue.

call pattern here:
```
   media_body = MediaIoBaseUpload(fd, mime_type[0])
   body = {
        'title': name,
        'mimeType': "application/vnd.google-apps.document",
    }
    service = _get_drive_service()
[...]

PR #406, opened 2017-06-14
MediaIoBaseDownload: Handle unknown media length

danielx:
Some media download APIs does not support partial downloads and do not return a 'Content-length' or 'Content-range' header and is therefore never considered done.

This change is to consider responses where the media length is unknown to be done.

This commit could potentially break media downloads where the API supports partial download and only returns a 'Content-length' header when the entire file has been served/read.

This commit fixes issue #15


YBadiss:
How is this a fix? Doesn't it mean that if I rely on the `done` flag, I'll ever only get the first chunk of the file?

danielx:
@YBadiss only if the server does not respond with a `content-length` or `content-range` header which AFAIK it does for all APIs supporting chunked downloads.

jonparrott:
Thanks, @danielx!

Issue #382, opened 2017-04-09
details missing in googleapiclient.errors.HttpError


pdknsk:
Run some code that makes an invalid request.

```Python
  File "/home/user/.local/lib/python2.7/site-packages/googleapiclient/http.py", line 840, in execute
    raise HttpError(resp, content, uri=self.uri)
googleapiclient.errors.HttpError: <HttpError 400 when requesting https://appengine.googleapis.com/v1/apps/<project>/services/default?alt=json returned "Invalid request.">
```
Now when you get the exception and its ```content``` the error becomes slightly more verbose.
```JSON
{
  "error": {
    "code": 400,
    "message": "Invalid request.",
    "errors": [
      {
        "message": "Invalid request.",
        "domain": "global",
        "reason": "badRequest"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```
Making the same request manually to the REST API returns a significantly more useful response.
```JSON
{
  "error": {
    "code": 400,
    "message": "Invalid request.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "field": "service.split.allocations[1]",
            "description": "Allocation must be a value in the range (0, 1]."
          },
          {
            "field": "service.split.allocations",
            "description": "Traffic allocations must sum to 1."
          }
        ]
      }
    ]
  }
}
```

jonparrott:
Seems like it might just be dropping the "details" section on the floor. I'm happy to review a patch to fix this, or we'll get to it when we can.

sonlac:
@jonparrott Thanks in advance for a review.

PR #412, opened 2017-07-08
Fix details missing in googleapiclient.errors.HttpError

sonlac:
For instance, the **googleapiclient.errors.HttpError** returns only the message and the status code. Adding the _error details_ on that would be useful and the error becomes slightly more verbose. In the response returned from the server, if the error body contains the **details** field, just show it together with the error message for more details. For example:

`<HttpError 400 when requesting http://example.org returned "country is required". Details: "error details">`

Resolves: #382

### DIFF
--- a/googleapiclient/errors.py
+++ b/googleapiclient/errors.py
@@ -46,6 +46,7 @@ class HttpError(Error):
         raise TypeError("HTTP content should be bytes")
     self.content = content
     self.uri = uri
+    self.error_details = ''
 
   def _get_reason(self):
     """Calculate the reason for the error from the response content."""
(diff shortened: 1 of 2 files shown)


jonparrott:
@sonlac thanks for sending this. I hate to be a bother, but could you update this to not change the indentation (or at least put the indentation change in its own commit)? It makes it extremely hard for me to review otherwise.

sonlac:
@jonparrott Sorry about that. I just tried to reformat code to PEP8 style. Thanks.

jonparrott:
Thanks, @sonlac.

Issue #314, opened 2016-11-28
missing requirements.txt


Pomax:
This example repo is missing the requirements.txt (which you can get from `pip freeze > requirements.txt` that tells us what needs to be installed

Issue #285, opened 2016-10-05
Please declare the version dependencies for your sample apps


nickretallack:
I am trying to run the [django sample app](https://github.com/google/google-api-python-client/tree/master/samples/django_sample) but I don't know what version of django and oauth2client it requires.  Running it with the latest does not work because these import paths no longer exist:

``` python
from django.core.management import execute_manager  # deprecated in django 1.4
from oauth2client.contrib.django_orm import FlowField   # possibly moved to django_utils?
```

PR #413, opened 2017-07-08
Fixing django-sample

otherpirate:
Issues: #285 #314 

Adding requirements.txt
Creating Makefile
Fixing problems with newer versions of oauth2
Fixing bugs in django files


jonparrott:
Thanks, @otherpirate!

PR #434, opened 2017-08-29
BatchHTTPRequest: only add access token header if we have it

### LINKED ISSUE #350: `TypeError` when trying to issue `BatchHttpRequest` when service was built using application default credentials

snarfed:
re-fixes #350. it was originally fixed by c669a3b9a89402c14179d11e587f5438b7209b5c following discussion in https://github.com/google/google-api-python-client/pull/351#discussion_r101218589 , but then that fix got lost in the next commit, https://github.com/google/google-api-python-client/pull/376/commits/282fc889956e42de8ec3e26aecbe39509f027333 .

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1207,7 +1207,7 @@ class BatchHttpRequest(object):
 
     if request.http is not None:
       credentials = _auth.get_credentials_from_http(request.http)
-      if credentials is not None:
+      if credentials is not None and getattr(credentials, 'access_token', None):
         _auth.apply_credentials(credentials, headers)
 
     # MIMENonMultipart adds its own Content-Type header.


jonparrott:
Can you show me a case where this is problematic?

snarfed:
sure! my repro is pretty much the same as @mgilson's in his original report in #350:

```py
from apiclient import discovery
import httplib2

service = discovery.build('plus', 'v1')
batch = service.new_batch_http_request()
batch.add(service.comments().list(activityId='123'))

# creds is an instance of oauth2client.client.OAuth2Credentials
http = creds.authorize(httplib2.Http())
batch.execute(http=http)
```

which results in:

```
  File "<console>", line 1, in <module>
  File ".../oauth2client/oauth2client/util.py", line 137, in positional_wrapper
  File ".../googleapiclient/http.py", line 1424, in execute
    self._execute(http, self._order, self._requests)
  File ".../googleapiclient/http.py", line 1340, in _execute
    body = self._serialize_request(request)
  File ".../googleapiclient/http.py", line 1211, in _serialize_request
    _auth.apply_credentials(credentials, headers)
  File ".../googleapiclient/_auth.py", line 113, in apply_credentials
    return credentials.apply(headers)
  File ".../oauth2client/oauth2client/client.py", line 621, in apply
    headers['Authorization'] = 'Bearer ' + self.access_token
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

with this PR, it runs fine.

now, to get it to pass the tests... :P

mgilson:
I don't know if it's helpful -- I haven't looked at this for quite a while, but way-way back I put together a branch/fork that seemed to work.  You can see my changes:  https://github.com/mgilson/google-api-python-client/compare/master...mgilson:app-default-creds?expand=1

Obviously don't trust that this actually works :-).

snarfed:
@mgilson it does work! i actually ran that patch in production for months, until just yesterday, and based this PR on it. thank you!

dtheodor:
Can you get this merged to finally fix the `TypeError: cannot concatenate 'str' and 'NoneType' objects` travesty?

jonparrott:
I don't think this PR covers the case where `google_auth` is used as the auth library instead of `oauth2client`, so I need to do some more investigation to see why this is occurring and how to properly fix it.

jonparrott:
Okay, getting to the source of this issue it's this:

1. `batch.execute()` allows you to specify an http object that will be used in place of all of the request's http objects.
1. When batch is serializing each request, the *credentials* attached to the http object for each request are preserved. 
1. The http object for each request is *never* preserved. If no `http` argument is passed to `batch.execute`, it just finds the first valid `http` client in the list of requests.

Because of this behavior, when you have the following situation:

1. You are using application default credentials for `discovery.build`
2. You are using different credentials in your `batch.execute`.

Then the `batch._serialize_request` tries to use the `discovery.build` credentials for each request instead of the credentials attached to to the `batch.execute` call.

I *can* remedy this by making `batch` always use the credentials specified by the http passed to `execute` and *ignore* the credentials attached the the request's http object. However, this prevents having different credentials for each request in a batch. I don't know if anyone is depending on that "feature", but I need to do some more investigation before I can commit one way or the other.

jonparrott:
I've decided I'm going to resolve this by making `apply_credentials` will proactively refresh the credentials if needed and will treat a `access_token` of `None` as a credential that needs refresh.

This does not in itself fix the example given. You will need to do one more thing to make this work as expected which is *explicitly* tell `discovery.build` *not* to use default credentials by specifying `http=httplib2.Http()` to the constructor.

PR incoming, but let me know if anyone disagrees with this.

snarfed:
thank you for working on this @jonparrott! that approach sgtm. i can easily add `http=httplib2.Http()` to my `discovery.build`. i'll happily try out your PR when it's pushed and see if it works for me.

jonparrott:
From my testing doing that alone fixes this in the `batch.execute(http)` case, so feel free to go ahead and try it (PR incoming, though)

PR #438, opened 2017-09-04
Allow http+credentials in build_from_document

e4p:
build_from_document was recently changed to make the passing of an HTTP object and credentials mutually exclusive. As far as I can see, there was no reason to do this. In the previous version one could pass either without any problems and with minor modifications that's true here as well.

This pull requests restores this functionality so that users relying on this ability can upgrade.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,21 @@ cache:
 matrix:
   include:
   - python: 2.7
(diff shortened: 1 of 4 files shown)


e4p:
withdrawing pull request until I debug this travis issue.

PR #440, opened 2017-09-05
Allow http+credentials in build_from_document.

e4p:
1) Prevent the passing of http and credentials to build_from_document
    from raising a ValueError.
2) End the behavior where credentials will be auto-created if none are
   provided. As far as I can tell, this new behavior is not documented
   anywhere and it breaks api compatibility between the 1.5 and 1.6
   releases. This is also preventing a migration I am working on involving
   a larger-ish code base that relies on the established behavior for testing.

### DIFF
--- a/googleapiclient/_auth.py
+++ b/googleapiclient/_auth.py
@@ -72,26 +72,29 @@ def with_scopes(credentials, scopes):
             return credentials
 
 
-def authorized_http(credentials):
+def authorized_http(credentials, http=None):
     """Returns an http client that is authorized with the given credentials.
 
     Args:
         credentials (Union[
             google.auth.credentials.Credentials,
             oauth2client.client.Credentials]): The credentials to use.
+        http (httplib2.Http): A httplib2 object or one similar enough to
+            provide authentication and tansport interfaces.
 
     Returns:
         Union[httplib2.Http, google_auth_httplib2.AuthorizedHttp]: An
             authorized http client.
     """
-    from googleapiclient.http import build_http
+    if http is None:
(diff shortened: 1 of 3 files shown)


e4p:
Looks like the queue for travis is pretty busy in the google org. For context, here's the travis run on the my project: https://travis-ci.org/e4p/google-api-python-client/builds/272156318

TLDR: tests are passing on Travis.

jonparrott:
I'm very concerned with merging this PR.

> Prevent the passing of http and credentials to build_from_document
from raising a ValueError.

Why do this? This was explicitly removed because it's *super* ambiguous and leads to really, really weird error cases. For example, a lot of old samples passed in an authorized http instance:

```
http = httplib2.Http
authed_http = credentials.authorize(http)
client = googleapiclient.discovery.build('storage', 'v1', http=http)
```

Passing the credentials as well creates a weird situation: does `discovery` wrap the http with the credentials? If it does, then the http instance is now *double wrapped*. If it doesn't, then what is it supposed to do with the credentials?

> End the behavior where credentials will be auto-created if none are
provided. As far as I can tell, this new behavior is not documented
anywhere and it breaks api compatibility between the 1.5 and 1.6
releases. This is also preventing a migration I am working on involving
a larger-ish code base that relies on the established behavior for testing.

This will break several cloud samples. Can we not retain this? Passing `credentials=False` retains the old behavior of creating a credentialless client. What is it about this old codebase that makes doing that untenable? Alternatively I could maybe see adding a `credentials=True` to opt-in to ADC.

e4p:
All fair points. Withdrawing

PR #446, opened 2017-10-01
Make body optional for requests with no parameters

### LINKED ISSUE #416: Dataproc API client fails to cancel due to wrong discovery of parameters
We (@adrianAhne and I) have found a small bug on the Python client code.

When using the Google API Python client to cancel a job, the Python code fails because it expects a body. However, the API documentation clearly states that there is no body to this request (https://cloud.google.com/dataproc/docs/reference/rest/v1beta2/projects.regions.jobs/cancel ). I do not know if the discovery process is wrong or it is something else.

Here is a simple excerpt to reproduce the problem:
```
from oauth2client.client import GoogleCredentials

credentials = GoogleCredentials.get_application_default()
service = build('dataproc', 'v1beta2', credentials=credentials)
request = service.projects().regions().jobs().cancel(
                  projectId='my_project_id',
                  region='global',  # cannot be anything else than global at the moment
                  jobId='my_job_id')
result = request.execute()
```

Which fails when creating the request with the following error:
```
TypeError: Missing required parameter "body"
```

There is a workaround, which is to remove the body from the request:
```
request = service.projects().regions().jobs().cancel(
                  projectId='my_project_id',
                  region='global',  # cannot be anything else than global at the moment
                  jobId='my_job_id',
                  body='')  # Set body to something, None does not work
request.body = None
request.body_size = 0
result = request.execute()
```

Maerig:
Fixes #416.
I made body optional instead of not adding it at all for backward compatibility.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -459,7 +459,7 @@ def _media_path_url_from_info(root_desc, path_url):
   }
 
 
-def _fix_up_parameters(method_desc, root_desc, http_method):
+def _fix_up_parameters(method_desc, root_desc, http_method, schema):
   """Updates parameters of an API method with values specific to this library.
 
   Specifically, adds whatever global parameters are specified by the API to the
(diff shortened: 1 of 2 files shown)


jonparrott:
Thanks, @Maerig!

Issue #450, opened 2017-10-25
Docstring refers to nonexistent class MediaDownloadStatus


kevinhb:
https://github.com/google/google-api-python-client/blob/af6766918a50ae3266037e1c30efe8cdacbc8f25/googleapiclient/http.py#L657

This method actually returns a MediaDownloadProgress object

bantini:
@kevinhb Thanks for pointing this out. I hope the merge solved your problem.

PR #451, opened 2017-10-27
Update Docstring (#1)

bantini:
Solves #450 
Updated the Docstring and corrected it for next_chunk()


jonparrott:
Thanks, @bantini

Issue #458, opened 2017-11-27
TypeError raised if returned error not in expected format


saicheems:
https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L115

^ We need a check for cases where 'reason' is not under the expected structure.

saicheems:
@bantini

bantini:
Also mentioned in #442 as well

bantini:
The latest commit should handle it.

Issue #442, opened 2017-09-14
Checking 403 reason assumes decoded data is a dict


stevedalton:
When checking for the 403 'reason' [here](https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L115), there's an implicit assumption the data is already a dictionary.  That's not always true: for example, here's some decoded data from the Google Monitoring API:

    [{
      u'error': {
        u'status': u'PERMISSION_DENIED', 
        u'code': 403, 
        u'errors': [{...}
    }]

This is (for whatever reason) a list, and so the code linked above fails with a TypeError (trying to index a list with a string), which is an uncaught exception.

bantini:
The latest commit should handle it.

PR #459, opened 2017-12-04
fix: 458 handle error format

bantini:
Addresses #442 and #458 

Handles cases where error object is a list instead of a dict.


bantini:
@jonparrott Do you think this is ready to merge?

jonparrott:
Yep, was just waiting on CI.

bantini:
Thank you.

PR #460, opened 2017-12-05
Add environment markers to django test dependency

jonparrott:


jonparrott:
/cc @lukesneeringer 

(Trivial test-only change, will merge once CI passes)

PR #464, opened 2018-01-11
Add convenience function to build subresource clients in one request

rvandegrift:
One of our apps needs to use googleapiclient to hit the GCP APIs for multiple projects, using different credentials.  This gets pretty repetitive with all of the subresources that the GCP APIs have:

```python
iam_a = build('iam', 'v1', credentials=proj_a)
iam_roles_a = iam_a.roles()
iam_b = build('iam', 'v1', credentials=proj_b)
iam_roles_b = iam_b.roles()

crm_a = build('cloudresourcemanager', 'v1beta1', credentials=proj_a)
crm_keys = crm_a.projects().serviceAccounts().keys()
```

After refactoring a bit, `googleapiclient.discovery` seems like the best place for this functionality.  With the change in this branch, the above becomes:
```
iam_roles_a = build_subresource('iam.roles', 'v1', credentials=proj_a)
iam_roles_b = build_subresource('iam.roles', 'v1', credentials=proj_b)
crm_a = build_subresource('cloudresourcemanager.projects.serviceAccounts.keys', 'v1beta1', credentials=proj_a)
```

Would you consider merging?

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -23,6 +23,7 @@ from six.moves import zip
 __author__ = 'jcgregorio@google.com (Joe Gregorio)'
 __all__ = [
     'build',
+    'build_subresource',
     'build_from_document',
     'fix_method_name',
     'key2param',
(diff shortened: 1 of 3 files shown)


rvandegrift:
I signed it!

jonparrott:
Hi @rvandegrift, I really appreciate the effort that went into contributing this patch. Unfortunately, this library is in maintenance mode so we're not adding any new features unless absolutely necessary.

This does seem really useful so maybe a blog post with this helper could be useful for users looking for a similar feature. :)

rvandegrift:
Hi @jonparrott - I figured I might get that response.  Is there another SDK out there that supports the complete GCP API?  google-cloud-python is very nice for what it supports.  But especially for the lower-level setup stuff, there's a lot that it cannot do

If the right answer is to submit feature requests there, then fair enough.

jonparrott:
Generally yes, `google-cloud-python` is the library where new features are actively being worked on.

PR #467, opened 2018-01-17
HttpRequest.__init__ remove useless mimeparse

### LINKED ISSUE #284: HttpRequest.__init__ useless mimeparse
Hi,
the code in the `HttpRequest.__init__` is useless:

``` python
    # Pull the multipart boundary out of the content-type header.
    major, minor, params = mimeparse.parse_mime_type(
        self.headers.get('content-type', 'application/json'))
```

And it prevents from `storage.insert` creating explicit dirs without content-type. I.e. this code pass only with the above code removed:

``` python
    api = googleapiclient.discovery.build('storage', 'v1', credentials=credentials)
    media = googleapiclient.http.MediaIoBaseUpload(io.BytesIO(b''), '')
    resp = api.objects().insert(bucket=bucket, name='testdir/', media_body=media).execute()
```

BTW similar code with empty contentType works on the .NET client, too.

xmedeko:
Fix #284

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -63,7 +63,6 @@ except ImportError:
   from oauth2client import _helpers as util
 
 from googleapiclient import _auth
-from googleapiclient import mimeparse
 from googleapiclient.errors import BatchError
 from googleapiclient.errors import HttpError
 from googleapiclient.errors import InvalidChunkSizeError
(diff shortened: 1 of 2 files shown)


jonparrott:
@xmedeko this might be useful, can you add a test that demonstrates the case?

xmedeko:
A test code is in the #284. Do you mean a test into the `test_http.py`?

jonparrott:
yes

xmedeko:
Added a simple test of `HttpRequest.__init__` by amend.
Note: the `mimeparse.parse_mime_type` is not used and may be removed?
Note2: `http.py` has some unused imports: `base64, gzip, sys`

jonparrott:
Thanks, @xmedeko

PR #471, opened 2018-01-19
Sheets API with Examples using google-auth

Barathwaja:
Added Sheets API using latest google-auth library since oauth2client is being deprecated.
This example uses Service Account (json file). It consist of two basic example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ Documentation for the YouTube Data API in
 Lets you view and manage and build smart apps using Google Spreadsheet.
 
 Documentation for the Sheets API in
-[PyDoc](https://developers.google.com/`s/api/quickstart/python).
+[PyDoc](https://developers.google.com/sheets/api/quickstart/python).
 
 <table>
   <tr>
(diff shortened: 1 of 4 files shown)


jonparrott:
Hi @Barathwaja, thank you for doing this but I don't think this is the appropriate place for these samples.

Barathwaja:
Hi @jonparrott , ok so where should I PR this then Coz you only mentioned that this is the correct place to PR. please refer the [PR](https://github.com/GoogleCloudPlatform/google-auth-library-python/pull/235)

grant:
Please create an issue before sending a PR.

There are lots of sheets examples here:
https://github.com/gsuitedevs/browser-samples

Barathwaja:
@jonparrott which you couldn't said directly.
@grant so I can submit an issue and send a PR in the link you gave? Mine is Python script connecting to Sheets API

grant:
@Barathwaja We already have a sample for python + Sheets, so I don't think we need a new one. I don't know what this PR is solving.

Please create an issue before a PR.

Barathwaja:
@grant plain and simple python libraries used by google is deprecated which google has said before and there is not a clear example of using python google sheet api with new libraries. And the documentation in Sheet API using python is outdated one.

PR #474, opened 2018-01-26
add path for client_secrets.json in sample_tools.py

t-Asai:
I add path for client_secrets.json in sample_tools.py, because we have some accounts then we need to prepare some client_secrets.json.
To use the same script with several json files, I made each directory and set env files in the directories.

### DIFF
--- a/googleapiclient/sample_tools.py
+++ b/googleapiclient/sample_tools.py
@@ -72,7 +72,7 @@ def init(argv, name, version, doc, filename, scope=None, parents=[], discovery_f
   # application, including client_id and client_secret, which are found
   # on the API Access tab on the Google APIs
   # Console <http://code.google.com/apis/console>.
-  client_secrets = os.path.join(os.path.dirname(filename),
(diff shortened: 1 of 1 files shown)


t-Asai:
I signed it!

googleapiclient/sample_tools.py
```diff
@@ -72,7 +72,7 @@ def init(argv, name, version, doc, filename, scope=None, parents=[], discovery_f
   # application, including client_id and client_secret, which are found
   # on the API Access tab on the Google APIs
   # Console <http://code.google.com/apis/console>.
-  client_secrets = os.path.join(os.path.dirname(filename),
+  client_secrets = os.path.join(os.path.dirname(filename), os.getcwd(),
```
jonparrott: I don't think this works the way you expect. What's the case that this is supposed to fix?

sduskis:
This PR is stale, and needs attention.  @t-Asai, do you want to follow up on comments, or should we close this?

busunkim96:
Closing this as the original author @t-Asai has not responded. Please feel free to re-open.

PR #486, opened 2018-03-16
Add Batch retry support

thehesiod:
This implements retry support for batched requests, both on the outer request, and the inner requests

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1436,9 +1436,6 @@ class BatchHttpRequest(object):
       redo_requests = {}
       redo_order = []
 
-      if retry_num >= num_retries:
-        break
-
       if retry_num > 0:
         # Sleep before retrying.
         sleep_time = self._rand() * 2 ** retry_num
@@ -1465,8 +1462,13 @@ class BatchHttpRequest(object):
       if redo_requests:
         self._execute(http, redo_order, redo_requests)
         retry_num += 1
-      else:
-        break
+
+        if retry_num >= num_retries:
+          break
+
+        continue
+
+      break
 
     # Now process all callbacks that are erroring, and raise an exception for
     # ones that return a non-2xx response? Or add extra parameter to callback
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -150,6 +150,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
   """
   resp = None
   content = None
+  exception = None
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
       # Sleep before retrying.
@@ -161,7 +162,6 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
       sleep(sleep_time)
 
     try:
-      exception = None
       resp, content = http.request(uri, method, *args, **kwargs)
     # Retry on SSL errors and socket timeout errors.
     except _ssl_SSLError as ssl_error:
@@ -1113,6 +1113,10 @@ class BatchHttpRequest(object):
     # A map of id(Credentials) that have been refreshed.
     self._refreshed_credentials = {}
 
+    # Stubs for testing.
+    self._rand = random.random
+    self._sleep = time.sleep
+
   def _refresh_and_apply_credentials(self, request, http):
     """Refresh the credentials and apply to the request.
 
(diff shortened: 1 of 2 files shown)


thehesiod:
I signed it!

thehesiod:
btw I noticed the files have a ton of PEP warnings, like files are 2 spaced instead of 4 spaced.  Not sure if you guys want these issues fixed too.

jonparrott:
Hey @thehesiod, thank you so much for putting this together. However, this library is in maintenance mode so unfortunately we won't be accepting new features.

I'm going to close this for now but I'm happy to continue discussing it. If we really feel like this is critical to the on-going maintenance of the library, we can reconsider.

thehesiod:
is there some replacement?

jonparrott:
For many cloud APIs active development is happening over at https://github.com/GoogleCloudPlatform/google-cloud-python. This library is still supported, just not receiving any new features. (New features = new bugs = instability, unfortunately).

nashley:
@jonparrott The bottom of the [google-cloud-python website](https://googlecloudplatform.github.io/google-cloud-python/) states the following:
> The Google APIs Python Client is a client library for using the broad set of Google APIs. google-cloud is built specifically for the Google Cloud Platform and is the recommended way to integrate Google Cloud APIs into your Python applications. If your application requires both Google Cloud Platform and other Google APIs, the 2 libraries may be used by your application.

Why deprecate (or put "in maintenance mode") this library, when its successor doesn't fully replace it (e.g., Google Drive isn't supported)?

Personally, I think having an in-library solution to exponential backoff and retries makes a lot of sense, since the practice is [strongly encouraged](https://developers.google.com/drive/v3/web/handle-errors#exponential-backoff) but non-trivial due to non-idempotent actions.
I understand not wanting to potentially introduce new bugs, but I'm surprised that this hasn't been in the library since 2015.

At the very least, having some documentation on the recommended way of implementing it through third-party libraries would be greatly appreciated. I've been using the tenacity library, and I'm having trouble identifying exactly which library calls need it and what exceptions it should retry on. Here's roughly what I've got so far:
```
import tenacity


def exception_warrants_retry(exception):
	"""Determine if the exception warrants (exponential backoff) retrying"""
	http_error_list = [
		403,  # Rate Limit Exceeded
		429,  # Too many requests / Rate Limit Exceeded (same as 403?)
		500   # Backend Error
	]

	if isinstance(exception, apiclient.errors.HttpError) and exception.resp in http_error_list:
		# Google suggests retrying after receiving this Drive API error (see https://developers.google.com/drive/v3/web/handle-errors)
		return True
	elif isinstance(exception, httplib2.ServerNotFoundError):
		# Can't connect to the remote host for some reason
		return True
	else:
		return False


@tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, max=10), retry=(tenacity.retry_if_exception(exception_warrants_retry)))
def method_that_uses_google_apis():
	pass
```

thehesiod:
@nashley ya it's pretty strange that Google is deprecating the only python library for a large set of APIs without any available replacement, although I've seen this behavior repeated in various other google products.  We're using this python library for calendar, admin, gmail, contacts, and a few other services.  We're almost to the point of writing our own async version of the API from the ground up.   Last thing we ran into is that auth in batch doesn't support retries and we're getting 5xx's all over the place from the Google endpoints (added to my PR).

jonparrott:
Let me be clear: this library is *not* deprecated and we are not planning on doing so unless there is a complete replacement. As such - we don't actually plan to have a complete replacement any time soon - except for Cloud APIs. This will still be the library for non-Cloud APIs. If we *do* deprecate it, we will still support it for at least a year as per our public deprecation policy.

All that said, it's **not** deprecated. We just are not adding new features. Yes, that can be frustrating and I hear you. However, adding new features also adds new bugs and introduces instability. We have a large majority of users who have used this library in code that they do not have desire to change, so even behavioral changes are problematic for them. We have to be extremely conservative here. "No" is temporary, but "yes" is forever. 

---

> Personally, I think having an in-library solution to exponential backoff and retries makes a lot of sense

We do on non-batch methods- you can call `execute(retry=5)`.

> At the very least, having some documentation on the recommended way of implementing it through third-party libraries would be greatly appreciated.

That's a wonderful Idea. I'd be more than happy to merge pull requests that add documentation around common patterns, workarounds, etc.

PR #493, opened 2018-04-20
Remove oauth2client._helpers dependency

helen-fornazier:
oauth2client is deprecated
Migrate required helper's code from oauth2client to googleapiclient to
decrease our dependency on oauth2client

_helpers.py copied from oauth2client code
https://github.com/google/oauth2client/blob/master/oauth2client/_helpers.py

### DIFF
--- a/googleapiclient/sample_tools.py
+++ b/googleapiclient/sample_tools.py
@@ -27,9 +27,13 @@ import os
 
 from googleapiclient import discovery
 from googleapiclient.http import build_http
-from oauth2client import client
-from oauth2client import file
-from oauth2client import tools
+
+try:
+    from oauth2client import client
+    from oauth2client import file
+    from oauth2client import tools
+except ImportError:
+    raise ImportError('googleapiclient.sample_tools requires oauth2client. Please install oauth2client and try again.')
 
 
 def init(argv, name, version, doc, filename, scope=None, parents=[], discovery_filename=None):
--- a/googleapiclient/_helpers.py
+++ b/googleapiclient/_helpers.py
@@ -1,0 +1,204 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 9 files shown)


helen-fornazier:
Please review, I am not entirely sure if this is the appropriated solution.
But with this change and #492 I can already manipulate instances using just google-auth, after uninstalling oauth2client

More specifically, what I am trying to run with google-auth instead of oauth2client is:
[code]: https://github.com/ClusterLabs/fence-agents/blob/master/fence/agents/gce/fence_gce.py
[PR]: https://github.com/ClusterLabs/fence-agents/pull/197

Thanks

helen-fornazier:
@jonparrott done, I imported the tests too as suggested. Could you please take a look in the commits?

Thanks

helen-fornazier:
Also, now the only file inside `googleapiclient/` who uses oauth2client is `googleapiclient/sample_tools.py`, do you know if this module is used by external libraries?
It seems to me that this file is just an utility module used by `samples/`, so I was wondering if we could move `sample_tools.py`to the `samples/` folder and completely remove oauth2client from the `install_requires` list in the setup.py

helen-fornazier:
@jonparrott lgtm, I added your suggestion in the PR, could you take a look please?

Can I remove oauth2client from the install_requires list or replace it by google-auth? https://github.com/google/google-api-python-client/blob/master/setup.py#L67

googleapiclient/sample_tools.py
```diff
@@ -27,9 +27,13 @@
...
 from googleapiclient.http import build_http
-from oauth2client import client
-from oauth2client import file
-from oauth2client import tools
+
+try:
+    from oauth2client import client
+    from oauth2client import file
+    from oauth2client import tools
+except:
```
jonparrott: Avoid broad excepts, use `except ImportError`.

jonparrott:
> Can I remove oauth2client from the install_requires list or replace it by google-auth? 

Yes, but let's two in two releases: one to have this, and then a second one to remove oauth2client.

googleapiclient/sample_tools.py
helen-fornazier: Fixed

jonparrott:
@helen-fornazier thank you so much for doing this!

PR #495, opened 2018-04-23
fix socket.timeout retry

thehesiod:
resolves https://github.com/google/google-api-python-client/issues/494


thehesiod:
ok updated, I think I can change the existing tests as there is no mention of TimeoutError being thrown for sockets, instead even for python2 `socket.timeout` is what it should be: https://docs.python.org/2/library/socket.html#socket.timeout

jonparrott:
@thehesiod if you rebase, #496 should fix the test failure.

PR #499, opened 2018-04-27
Drop oauth2client dependency 

### LINKED ISSUE #437: Deprecation of oauth2client dependency
According to the readme google/oauth2client is deprecated. They suggest switching to [google-auth](https://google-auth.readthedocs.io/) or [oauthlib](http://oauthlib.readthedocs.io/). 

This probably means that this package and all of the docs should also be refactored into using these new packages?

helen-fornazier:
oauth2client is deprecated, use google-auth instead
Closes #437

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,41 @@ cache:
 matrix:
   include:
   - python: 2.7
-    env: TOXENV=py27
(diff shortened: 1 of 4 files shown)


amureki:
Thank you, everyone, for finalizing this! ✨ ✨ ✨

PR #504, opened 2018-05-09
Update groupsettings.py

charlesreid1:


mcdonc:
@charlesreid1 thanks a lot for the PR!  Can you add a short description about the problem it's solving?

charlesreid1:
This fixes an import issue with oauth2client. In oauth2client.tools there is no `run`, only `run_flow`: https://github.com/google/oauth2client/blob/2da8ccde3f74507990cb551fe48f6d25820e6ab3/oauth2client/tools.py#L37

mcdonc:
Thanks @charlesreid1 !

Issue #524, opened 2018-06-04
oauth2client removed from setup.py deps but still imported


epall:
Building on #521, dropping `oauth2client` from `setup.py` in #499 is causing problems importing `apiclient.discovery`:

```
Python 3.6.5 (default, Mar 31 2018, 13:05:03)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from apiclient import discovery
Traceback (most recent call last):
  File "/Users/eric/projects/google-api-python-client/googleapiclient/sample_tools.py", line 32, in <module>
    from oauth2client import client
ModuleNotFoundError: No module named 'oauth2client'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/eric/projects/google-api-python-client/apiclient/__init__.py", line 13, in <module>
    from googleapiclient import sample_tools
  File "/Users/eric/projects/google-api-python-client/googleapiclient/sample_tools.py", line 36, in <module>
    raise ImportError('googleapiclient.sample_tools requires oauth2client. Please install oauth2client and try again.')
ImportError: googleapiclient.sample_tools requires oauth2client. Please install oauth2client and try again.
```

```
$ git grep oauth2client | grep -v -e docs\/ -e CHANGELO
.travis.yml:    env: TOXENV=py27-oauth2client1
.travis.yml:    env: TOXENV=py27-oauth2client2
.travis.yml:    env: TOXENV=py27-oauth2client3
.travis.yml:    env: TOXENV=py27-oauth2client4
.travis.yml:    env: TOXENV=py34-oauth2client1
.travis.yml:    env: TOXENV=py34-oauth2client2
.travis.yml:    env: TOXENV=py34-oauth2client3
.travis.yml:    env: TOXENV=py34-oauth2client4
.travis.yml:    env: TOXENV=py35-oauth2client1
.travis.yml:    env: TOXENV=py35-oauth2client2
.travis.yml:    env: TOXENV=py35-oauth2client3
.travis.yml:    env: TOXENV=py35-oauth2client4
.travis.yml:    env: TOXENV=py36-oauth2client1
.travis.yml:    env: TOXENV=py36-oauth2client2
.travis.yml:    env: TOXENV=py36-oauth2client3
.travis.yml:    env: TOXENV=py36-oauth2client4
[...]

PR #525, opened 2018-06-04
Make apiclient.sample_tools optional

epall:
It depends heavily on oauth2client, which has been dropped as a required dependency

This seems to me like the quickest way to tackle #524, but it feels ugly. Maybe `sample_tools` needs to get re-written to use `google-auth`?

### DIFF
--- a/apiclient/__init__.py
+++ b/apiclient/__init__.py
@@ -10,7 +10,12 @@ from googleapiclient import errors
 from googleapiclient import http
 from googleapiclient import mimeparse
 from googleapiclient import model
-from googleapiclient import sample_tools
+try:
+    from googleapiclient import sample_tools
+except ImportError:
+    # Silently ignore, because the vast majority of consumers won't use it and
+    # it has deep dependence on oauth2client, an optional dependency.
+    sample_tools = None
 from googleapiclient import schema
 
 __version__ = googleapiclient.__version__


apiclient/__init__.py
```diff
@@ -10,7 +10,11 @@
 from googleapiclient import http
 from googleapiclient import mimeparse
 from googleapiclient import model
-from googleapiclient import sample_tools
+try:
+    from googleapiclient import sample_tools
+except ImportError:
+    import sys
+    sys.stderr.write("""Warning: unable to import apiclient.sample_tools\n""")
```
theacodes: Don't write to stderr, just ignore it. It'll give a standard import error if someone tries to `from apiclient import sample_tools`, and I'm with that (as they should be using `from googleapiclient import sample_tools`.

apiclient/__init__.py
```diff
@@ -22,10 +26,14 @@
     'http': http,
     'mimeparse': mimeparse,
     'model': model,
-    'sample_tools': sample_tools,
     'schema': schema,
 }
 
+try:
```
theacodes: just do `sample_tools = None` in the try..except above and you won't need to modify this at all.

tseaver:
Should there be a `sample_tools` extra added in `setup.py`, with a dep on `oauth2client` for now?

PR #545, opened 2018-07-17
Fix #164 under Python 3; prevent line-folding at beginning of content-id header

### LINKED ISSUE #164: Content-ID header over 64 characters gets an extra newline; breaks batch requests
I'm working with batch requests, and adding custom request_id values. With a request_id value of length 28, the total length of the Content-ID value inside the `<` and `>` characters is 64.

Somehow, for >=64 characters, `"\n "` (including trailing space) gets added before the Content-ID value, but after the `"Content-ID: "`, resulting in a malformed Content-ID value which makes the round trip back (response takes the form `Content-ID: response- <uuid+request_id>` instead of `Content-ID: <response-uuid+request_id>`) and throws an exception (a BatchError for a malformed Content-ID value).

I've narrowed this down to this patch of googleapiclient/http.py starting at line 1234:

```
# encode the body: note that we can't use `as_string`, because
# it plays games with `From ` lines.
print(message) #no newline here
fp = StringIO()
g = Generator(fp, mangle_from_=False)
g.flatten(message, unixfrom=False)
body = fp.getvalue()
print(body,[body]) #newline appears here, printing a list shows newline as \n
```

This can be reproduced without requiring an authorised call, using the following test code:

```
from httplib2 import Http
directory = build('admin', 'directory_v1')

def errorTest(id_length):
    req_id = "abcdefghijklmnopqrstuvwxyz0123456789"
    try:
        batch_dir = directory.new_batch_http_request()
        batch_dir.add(directory.groups().list(), request_id=req_id[0:id_length])
        batch_dir.execute(http=Http()) #BatchError, unprintable (separate issue)
    except Exception as e:
        print(e.reason)
        raise

for i in [27,28]:
    print(i,"############################################",i) #readability
    try:
        errorTest(i)
    except AttributeError:
        #only occurs if call not authorised; BatchError occurs first and isn't caught
        pass
```

### LINKED ISSUE #536: Batch request for group members results in <BatchError "Invalid value for Content-ID: ...
Occurs in line 1191 of `http.py` - exact message is:

`googleapiclient.errors.BatchError: <BatchError "Invalid value for Content-ID: response- <cf4d6eb8-8f95-4fe1-ad82-2252fa9e8197+my-group@mydomain.com>">

If I insert `header = header.replace('response- ','')` at line 1175 the issue goes away.

mcdonc:
Closes #164 

Closes #536

Also ensures that things don't break if a user supplies a request_id with a "+" in it.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1171,7 +1171,10 @@ class BatchHttpRequest(object):
     if self._base_id is None:
       self._base_id = uuid.uuid4()
 
-    return '<%s+%s>' % (self._base_id, quote(id_))
+    # NB: we intentionally leave whitespace between base/id and '+', so RFC2822
+    # line folding works properly on Python 3; see
+    # https://github.com/google/google-api-python-client/issues/164
+    return '<%s + %s>' % (self._base_id, quote(id_))
 
   def _header_to_id(self, header):
     """Convert a Content-ID header value to an id.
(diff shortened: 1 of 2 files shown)


mcdonc:
@tseaver I figured out how to test it and added a test, could you re-review?

PR #548, opened 2018-07-19
Create results list for requests with no callback

chrish10e:
This PR will collect the results in a  `[ [response, exception] ]` list and return that list synchronously when no callbacks were provided. Allows batch execution to happen within the scope of a function rather than forcing out into a callback function.

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1462,7 +1462,8 @@ class BatchHttpRequest(object):
     # Now process all callbacks that are erroring, and raise an exception for
     # ones that return a non-2xx response? Or add extra parameter to callback
     # that contains an HttpError?
-
+    results = []
+  
     for request_id in self._order:
       resp, content = self._responses[request_id]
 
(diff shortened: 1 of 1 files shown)


mcdonc:
The code itself looks fine to me, but the change would require a docs update.  And given that it would require a docs update, I'm a bit -0 on the change, because people can supply a callback that does what this code does (append the response and exception to a list).  Also, it would require the addition of a few tests.

chrish10e:
Fair enough, but it doesn't feel very Python-like to force this async behavior, particularly given that the underlying process is happening synchronously. Maybe I'm missing something in the way this function is intended to work, but in my code at least, this batch call happens in the middle of a function, and pushing the response handling out into a separate function means I lose all context within the scope of my calling function.

mcdonc:
Use a function defined within the function you're using to call batch.execute() (ie. a closure):

```
def my_calling_func():
  
  L = []
  
  def append_to_list(request_id, response, exception):
    L.append([response, exception])

  batch = service.new_batch_http_request()

  batch.add(service.someresource().somemethod(), callback=append_to_list)
  batch.execute(http=Http())
```

PR #560, opened 2018-08-20
Add python3 support, pretty print table

jkowa:
print_table function now has support for python 3+ and cleaner output


jkowa:
I signed it!

parthea:
This PR appears to be obselete. I see the `print_table ` method was fixed in a subsequent PR #722.

PR #564, opened 2018-08-28
Fix HttpRequest json dump

vianney-g:
Response callbacks are not JSON serializable and should be excluded from
json dump.


vianney-g:
I signed it!

PR #585, opened 2018-11-09
Add a limit for the size of batch request

tristan-lin:
We have a limit of batch request which is defined in the doc[1].
However, in the batch class there is no such limit, which usually result
a timeout somewhere and very hard to debug.

[1] https://developers.google.com/api-client-library/python/guide/batch

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -73,6 +73,8 @@ DEFAULT_CHUNK_SIZE = 100*1024*1024
 
 MAX_URI_LENGTH = 2048
 
+MAX_BATCH_LIMIT = 1000
+
 _TOO_MANY_REQUESTS = 429
 
 DEFAULT_HTTP_TIMEOUT_SEC = 60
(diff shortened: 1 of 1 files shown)


theacodes:
Please add a corresponding test for this.

tristan-lin:
Done :)

tests/test_http.py
```diff
@@ -1077,6 +1077,7 @@ class TestBatch(unittest.TestCase):
 
   def setUp(self):
     model = JsonModel()
+    self.MAX_BATCH_LIMIT = 1000
```
tseaver: No point in adding an attribute in `setUp` to be used only in one testcase:  just import the constant in that method (see suggestion below).

tests/test_http.py
```diff
@@ -1192,6 +1193,19 @@ def test_add(self):
     batch.add(self.request1, request_id='1')
     self.assertRaises(KeyError, batch.add, self.request1, request_id='1')
 
+  def test_add_fail_for_over_limit(self):
```
tseaver:
```suggestion
  def test_add_fail_for_over_limit(self):
        from googleapiclient.http import MAX_BATCH_LIMIT

```

tests/test_http.py
```diff
@@ -1192,6 +1193,19 @@ def test_add(self):
...
+    batch = BatchHttpRequest()
+    for i in xrange(0, self.MAX_BATCH_LIMIT):
```
tseaver:
```suggestion
    for i in xrange(0, MAX_BATCH_LIMIT):
```

tristan-lin: Thanks. Done.

tests/test_http.py
tristan-lin: Thanks for the suggestion! Committed the change.

tristan-lin:
@tseaver thanks for your approval! 
@theacodes what else I could do to get the PR merged?

PR #600, opened 2018-12-07
Make http.MediaFileUpload close its file descriptor.

### LINKED ISSUE #575: Receive warnings "ResourceWarning: unclosed file" when using googleapiclient.http.MediaFileUpload
On google-api-python-client/googleapiclient/http.py:552, a file is opened. But this file is never closed. So I receive the following warning:

ResourceWarning: unclosed file <_io.BufferedReader name=<file_name> media_body = http.MediaFileUpload()

wangxf123456:
Fixes #575

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -551,14 +551,14 @@ class MediaFileUpload(MediaIoBaseUpload):
         in a single request.
     """
     self._filename = filename
-    self._fd = open(self._filename, 'rb')
+    fd = open(self._filename, 'rb')
     if mimetype is None:
       # No mimetype provided, make a guess.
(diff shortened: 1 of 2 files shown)


googleapiclient/http.py
```diff
@@ -551,16 +551,20 @@ def __init__(self, filename, mimetype=None, chunksize=DEFAULT_CHUNK_SIZE,
         in a single request.
     """
     self._filename = filename
-    fd = open(self._filename, 'rb')
+    self._fd = open(self._filename, 'rb')
```
tseaver: `self._fd` is already set in `MediaIoBaseUpload`.

PR #601, opened 2018-12-07
Change xrange to range

busunkim96:
`xrange` does not exist in Python 3 and this is causing Python 3 tests to fail.

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -1196,7 +1196,7 @@ class TestBatch(unittest.TestCase):
     from googleapiclient.http import MAX_BATCH_LIMIT
 
     batch = BatchHttpRequest()
-    for i in xrange(0, MAX_BATCH_LIMIT):
+    for i in range(0, MAX_BATCH_LIMIT):
       batch.add(HttpRequest(
         None,
         None,


tests/test_http.py
```diff
@@ -1196,7 +1196,7 @@ def test_add_fail_for_over_limit(self):
     from googleapiclient.http import MAX_BATCH_LIMIT
 
     batch = BatchHttpRequest()
-    for i in xrange(0, MAX_BATCH_LIMIT):
+    for i in range(0, MAX_BATCH_LIMIT):
```
tseaver: Might be cleaner to use `six.moves.range` here, e.g. via the following at the top of the module:

```python
from six.moves import range
```

tests/test_http.py
```diff
@@ -19,7 +19,7 @@
 Unit tests for the googleapiclient.http.
 """
 from __future__ import absolute_import
-from six.moves import range
+from six.moves import range, xrange
```
tseaver: Oops, I missed that the test module was already importing `range`.  You can back out that second commit.

busunkim96: Done

PR #613, opened 2019-01-15
Alias unittest2 import as unittest in test__auth.py.

janlugt:


janlugt:
This makes the way we import unittest consistent with the other tests.

tseaver:
Thanks for the patch, @janlugt!

PR #661, opened 2019-05-17
Convert Changelog to Markdown for releasetool.

busunkim96:
Also append a `##` to each version section, since this is how Releasetool knows where to insert the new changes.


busunkim96:
Obsolete

PR #696, opened 2019-06-14
Regenerate docs.

busunkim96:
I installed epydoc and ran [`docs/build`](https://github.com/googleapis/google-api-python-client/blob/master/docs/build).

Are there other steps to generate this?


busunkim96:
Yup, I ran the script. https://github.com/googleapis/google-api-python-client/blob/master/docs/build

Just wanted to check if there was more magic somewhere.

PR #699, opened 2019-06-14
docs: make setup.py PyPI friendly.

### LINKED ISSUE #698: Pypi Project Description
The Pypi package has a description that is very old and not useful.

> The Google API Client for Python is a client library for accessing the Plus, Moderator, and many other Google APIs.

https://pypi.org/project/google-api-python-client/1.7.9/

For example, the G+ API is deprecated.

This should be updated to something more useful. It's hard to tell that this is the real pypi package!

busunkim96:
README contents will be displayed on PyPI. Copied from https://github.com/googleapis/google-cloud-python/blob/master/speech/setup.py.

Closes #698 (once a new release is published).


busunkim96:
Obsolete, see #852

Issue #693, opened 2019-06-11
Documentation not accessible.


jerrythomas:
The documentation links are all redirected to this repository. The following links are from the readme 

* [Developer Documentation](https://developers.google.com/api-client-library/python/start/get_started)
* [full documentation for this library](https://developers.google.com/api-client-library/python/)

All of them redirect to https://github.com/googleapis/google-api-python-client
Even the links for the documentation on [Google API client libraries](https://developers.google.com/api-client-library/) redirects to this repo. 
* https://developers.google.com/api-client-library/python/
* https://developers.google.com/api-client-library/python/start/get_started
* https://developers.google.com/api-client-library/python/samples/samples

#### Steps to reproduce

  1. Click on the links in the readme. All of them redirect to this repo.

Samples are actually here
https://github.com/googleapis/google-api-python-client/tree/master/samples

I couldn't find documentation similar to the one for [java](https://developers.google.com/api-client-library/java/). Can you correct the links so that the documentation is accessible?

busunkim96:
Hi @jerrythomas, we're in the process of removing documentation on that site, and have open issues on this repo to add documentation here.

Thank you for filing an issue!

doransmestad:
Any recommendations in the meanwhile?  I'd like to use this project to interact with the Google APIs, but no available documentation makes this a bit challenging.  Is there a place we can find the documentation while it is being migrated?   Thanks

busunkim96:
Hi @doransmestad, if you're looking for the generated ref docs they are currently available under:
`https://googleapis.github.io/google-api-python-client/docs/dyn/{api_name}_{version}`

*Examples:
https://googleapis.github.io/google-api-python-client/docs/dyn/accesscontextmanager_v1beta
https://googleapis.github.io/google-api-python-client/docs/dyn/customsearch_v1

For other docs (auth guide, for example), there are currently PRs open on this repo that have the contents of those.

We know this isn't ideal, and we're working on getting the GH pages site up to date.

Thank you all for your patience!

csudcy:
Hey @busunkim96, I had a look through PRs but couldn't see any which look like they're for adding docs; it looks like there are some recent merges which update docs though so maybe the links in the root readme just need updating?

@doransmestad in the meantime, I've found this sample most helpful: https://github.com/googleapis/google-api-python-client/blob/master/samples/appengine/

busunkim96:
[...]

PR #702, opened 2019-06-19
Create index file for dynamically generated docs.

busunkim96:
Will close part of #693.


describe.py
```diff
@@ -394,6 +394,7 @@ def document_api_from_discovery_document(uri):
   if FLAGS.discovery_uri:
     document_api_from_discovery_document(FLAGS.discovery_uri)
   else:
+    api_directory = {}
```
tseaver: This could be simplified using `colletions.defaultdict`, e.g.:

```
import collections
...
    api_directory = collecctions.defaultdict(list)
```

and then later just:

```python
    api_directory[api['name']].append(api['version'])

File: docs/dyn/index.html [hunk 2]
```diff
@@ -0,0 +1,801 @@
...
+}
+
+.details {
+  font-weight: bold;
+  font-size: 14px;
+}
+
+</style>
+
+<h2>websecurityscanner</h2>
```
tseaver: Should we be sorting by API name here?  Maybe also by version within an API?

busunkim96: I thought they'd land in the order of the discovery doc, but I guess that was not the case. :upside_down_face:

PR #718, opened 2019-07-16
Never require 'body' for requests.

### LINKED ISSUE #713: cloudresourcemanager.project.getIamPolicy recently started throwing 'Missing required parameter "body"'
Replicated in version 1.7.7, 1.7.8, and 1.7.9 on MacOS, and various flavors of linux with python 3.7

Prior to July 3rd, the following code worked, and returned the IAM policy for a given project:

```
service = build('cloudresourcemanager','v1')
req = service.projects().getIamPolicy(resource='my-project')
resp = req.execute()
print(resp)
```

Since then, our existing applications using that functionality started throwing an error that the required parameter `body` was missing. At first I thought google made a breaking change to the discovery document, but upon checking, the discovery doc does not require the `body` parameter. Then I started poking around and I see this library adds it as required here: https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery.py#L494

All of the code related to this, that I noticed, is over 7 years old. So its unlikely that a change to the library broke anything. One thing that COULD explain the issue is a few lines below the line i linked above... IF the discovery document didn't previously send a `request` field in the method description, this library would make the body optional. I have no way to determine if that actually happened, as I don't have any older cached discovery documents for the cloudresourcemanager v1 API.

I realize what I'm reporting here isn't a bug with this library, but something appears to have changed the behavior in this case (and potentially others) so I wanted to:

-  report it (so others can find it)
-  see if you have a way to determine what happened

Thanks!

busunkim96:
Fixes #713. 

Never require that a `body` be passed in a request. Validation of `body` will be left to the server.

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -96,7 +96,6 @@ _MEDIA_SIZE_BIT_SHIFTS = {'KB': 10, 'MB': 20, 'GB': 30, 'TB': 40}
 BODY_PARAMETER_DEFAULT_VALUE = {
     'description': 'The request body.',
     'type': 'object',
-    'required': True,
 }
 MEDIA_BODY_PARAMETER_DEFAULT_VALUE = {
     'description': ('The filename of the media request body, or an instance '
@@ -494,9 +493,6 @@ def _fix_up_parameters(method_desc, root_desc, http_method, schema):
   if http_method in HTTP_PAYLOAD_METHODS and 'request' in method_desc:
     body = BODY_PARAMETER_DEFAULT_VALUE.copy()
     body.update(method_desc['request'])
-    # Make body optional for requests with no parameters.
-    if not _methodProperties(method_desc, schema, 'request'):
-      body['required'] = False
     parameters['body'] = body
 
   return parameters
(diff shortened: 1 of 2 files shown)


busunkim96:
If you forget to pass a body to an API that has required properties, the server will throw an error.

**Example**:

(`create` requires a body with a new job as its payload.)
```

service = build('jobs', 'v2')
req = service.jobs().create()
resp = req.execute()
```

```
    raise HttpError(resp, content, uri=self.uri)
googleapiclient.errors.HttpError: <HttpError 400 when requesting https://jobs.googleapis.com/v2/jobs?alt=json returned "Following fields are missing or invalid: Company Name, Description, Title, Job Requisition ID, One of Application Method (URL/Emails/Instructions). (Possible reasons: non-numeric for double field, invalid language code, etc.). Request ID for tracking:
```

TxBillbr:
How long will it take before this is visible to the client lib? Passing body={} in a request is very dangerous Python behavior. This broke a bunch of our automation.

busunkim96:
@TxBillbr I'll cut a release today -- apologies for the disruption this caused!

TxBillbr:
Thanks much for the prompt handling. I have just released a local work-around for our local library that tries the correct method and if it throws and error tries with the body={}.

busunkim96:
@TxBillbr This is available in the latest release [1.7.10](https://pypi.org/project/google-api-python-client/1.7.10/)

PR #721, opened 2019-07-21
tox.ini: Look for Python syntax errors and undefined names

cclauss:


tseaver:
Hmm, we don't have `tox` running on CI?

busunkim96:
The Kokoro job [runs tox](https://github.com/googleapis/google-api-python-client/blob/master/.kokoro/build.sh)  -- I made the mistake of forgetting to add `kokoro` label to make the CI run here. Fixed lint errors as part of #724

PR #722, opened 2019-07-21
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.  Also fixed some _undefined names_.

Four _undefined names_ remain...

$ __flake8 . --count --show-source --statistics --select=E9,F63,F7,F82__
```
./samples/coordinate/coordinate.py:60:13: F821 undefined name 'build'
  service = build('coordinate', 'v1', http=http)
            ^
./samples/coordinate/coordinate.py:60:44: F821 undefined name 'http'
  service = build('coordinate', 'v1', http=http)
                                           ^
./samples/coordinate/coordinate.py:64:46: F821 undefined name 'FLAGS'
    jobs_result = service.jobs().list(teamId=FLAGS.teamId).execute(http=http)
                                             ^
./samples/coordinate/coordinate.py:64:73: F821 undefined name 'http'
    jobs_result = service.jobs().list(teamId=FLAGS.teamId).execute(http=http)
                                                                        ^
4     F821 undefined name 'build'
4
```


googleapiclient/http.py
```diff
@@ -156,7 +156,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
       LOGGER.warning(
           'Sleeping %.2f seconds before retry %d of %d for %s: %s %s, after %s',
           sleep_time, retry_num, num_retries, req_type, method, uri,
-          resp.status if resp else exception)
+          resp.status if resp else Exception)
```
tseaver: This change loses the re-bound `exception` below, which means it loses information about the actual cause.

PR #724, opened 2019-07-22
Fix http.py: Exception -> exception.

busunkim96:
Fixes bug introduced in #722

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -149,6 +149,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
   """
   resp = None
   content = None
+  exception = None
   for retry_num in range(num_retries + 1):
     if retry_num > 0:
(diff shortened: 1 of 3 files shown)


googleapiclient/http.py
```diff
@@ -156,7 +156,7 @@ def _retry_request(http, num_retries, req_type, sleep, rand, uri, method, *args,
       LOGGER.warning(
           'Sleeping %.2f seconds before retry %d of %d for %s: %s %s, after %s',
           sleep_time, retry_num, num_retries, req_type, method, uri,
-          resp.status if resp else Exception)
+          resp.status if resp else exception)
```
tseaver: Should we assign `exception = None` above to prevent the "mark one eyeball linter" from firing on this line again?

busunkim96: Does this look correct? @tseaver

PR #742, opened 2019-08-13
Fix TypeError: search_analytics_api_sample.py #732

alvaroseparovich:
Solution given by thedataboi


alvaroseparovich:
@googlebot I signed it!

PR #750, opened 2019-08-20
Correct response access

vedgar:


vedgar:
@googlebot I signed it!

Honestly, the amount of bureaucracy I have to go through just to fix a simple typo is staggering. You really need a reality check IMO.

busunkim96:
@vedgar The CLA only needs to be signed once for all the google repos. :grin: Thanks!

PR #752, opened 2019-08-28
pip install -r requirements.txt

acmcoins:


acmcoins:
By mistake used online versions. Was trying to install python on my Google
store

PR #759, opened 2019-09-17
Update authorized_http()

vachan-shetty:
Update the authorized_http() method to accept a httplib2.Http parameter.

### DIFF
--- a/googleapiclient/_auth.py
+++ b/googleapiclient/_auth.py
@@ -76,19 +76,24 @@ def with_scopes(credentials, scopes):
             return credentials
 
 
-def authorized_http(credentials):
+def authorized_http(credentials, default_http=None):
     """Returns an http client that is authorized with the given credentials.
 
     Args:
         credentials (Union[
             google.auth.credentials.Credentials,
             oauth2client.client.Credentials]): The credentials to use.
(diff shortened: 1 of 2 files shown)


busunkim96:
Hi @vachan-shetty, thank you for submitting a PR!

Would you be able to provide some context on what issue this is addressing?

vachan-shetty:
Hi @busunkim96 ,
The authorized_http() method currently accepts only a credentials object as a parameter and uses the default Http object (created via build_http()) to return an authorized http client. 

The change I have made allows the method to create an authorized http client using an Http object supplied as a parameter OR defaults to the default Http object in case on is not provided. This allows the user to use a Http object of their choice to create the authorized http client.

busunkim96:
@vachan-shetty, you can already pass a custom `httplib2.Http` object to `discovery.build`, which calls `discovery.build_from_document.` Does that meet your needs? 

https://github.com/googleapis/google-api-python-client/blob/48b93fecd15f1eb181ad36bcd53335af7f88b301/googleapiclient/discovery.py#L168-L188
https://github.com/googleapis/google-api-python-client/blob/48b93fecd15f1eb181ad36bcd53335af7f88b301/googleapiclient/discovery.py#L350-L377

PR #762, opened 2019-10-07
Update admin_directory_v1.members.html

waltersdmh:
body['email'] field is not read-only. It is used to identify who to add to a group.


waltersdmh:
@googlebot I signed it!

busunkim96:
Hi @waltersdmh, thank you for submitting a PR!

The API-specific documentation in this repo is generated from the [discovery document](https://www.googleapis.com/discovery/v1/apis/admin/directory_v1/rest).

From the API docs [here](https://developers.google.com/admin-sdk/directory/v1/reference/members), an email cannot be changed after a user is inserted, so it is not a 'writable' field.

If you have suggestions on better ways to communicate that in the discovery document we would be open to suggestions.

PR #768, opened 2019-10-22
docs: recommend install with virtualenv

busunkim96:
Copied from google-cloud-python install docs: https://github.com/googleapis/google-cloud-python/blob/master/storage/README.rst

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,16 +16,30 @@ For Google Cloud Platform APIs such as Datastore, Cloud Storage or Pub/Sub, we r
 
 ## Installation
 
-To install, simply use `pip` or `easy_install`:
+Install this library in a [virtualenv](https://virtualenv.pypa.io/en/latest/) using pip. virtualenv is a tool to
+create isolated Python environments. The basic problem it addresses is one of
+dependencies and versions, and indirectly permissions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,16 +16,30 @@ For Google Cloud Platform APIs such as Datastore, Cloud Storage or Pub/Sub, we r
...
 ## Installation
 
-To install, simply use `pip` or `easy_install`:
+Install this library in a [virtualenv](https://virtualenv.pypa.io/en/latest/) using pip. virtualenv is a tool to
+create isolated Python environments. The basic problem it addresses is one of
+dependencies and versions, and indirectly permissions.
+
+With virtualenv, it's possible to install this library without needing system
+install permissions, and without clashing with the installed system
+dependencies.
```
tseaver: Hmmm, this is a little weird, as the `virtualenv` library really isn't needed for supported Python 3.x versions:  `python -m venv` works fine for them.  It is a "works everywhere" recommendation as long as we are supporting Python 2.7, though.

busunkim96: Opened #769 to track that the wording should be updated when Python 2.7 support is dropped

PR #770, opened 2019-10-24
chore(docs): fix capitalization in docs/README.md

prabhupant:
Library was spelt `LIbrary` with both L and I capitalized


busunkim96:
Thank you for the fix @prabhupant!

PR #771, opened 2019-10-29
docs: publish to googleapisdotdev, use kokoro templates

busunkim96:
TODO
- [x] open separate PR to run `black`

I attempted to be conservative with the docstring changes. Within a single file the style should be consistent.


busunkim96:
This is very stale now, closing.

PR #773, opened 2019-10-31
Update pythonapp.yml

tu-trankilo:
>


busunkim96:
Hi @tu-trankilo!

Thank you for the PR. We don't currently use Workflows for our CI. Was there a particular concern you wanted to address?

PR #813, opened 2020-01-24
fix: exclude 308s from httplib2 redirect codes list

### LINKED ISSUE #803: httplib2 v0.16.0 breaks the library
Recently updated httplib2 library (v0.16.0) have a breaking change causing this Exception:

`httplib2.RedirectMissingLocation: Redirected but response is missing a location: header`

**Steps to reproduce:**
1. `pip install --upgrade httplib2`
2. Now use the google drive library to upload a file (This is where I encountered the bug, it may break while using other stuffs a well)

It should either force to use 0.15.0 version of httplib2 or adapt the breaking changes introduced in the new version. 

A temporary fix for those using the client library would be to add httplib2==0.15.0 in requirements,txt of the project

busunkim96:
Closes #803


googleapiclient/http.py
```diff
@@ -1886,4 +1886,11 @@ def build_http():
         http_timeout = socket.getdefaulttimeout()
     else:
         http_timeout = DEFAULT_HTTP_TIMEOUT_SEC
-    return httplib2.Http(timeout=http_timeout)
+    http = httplib2.Http(timeout=http_timeout)
+    # 308's are used by several Google APIs (Drive, YouTube)
+    # for Resumable Uploads rather than Permanent Redirects.
+    # This asks httplib2 to exclude 308s from the status codes
+    # it treats as redirects
+    http.redirect_codes = set(http.redirect_codes) - {308}
```
plamut: Let's cover this change with another test case in `TestHttpBuild`.

googleapiclient/http.py
```diff
@@ -1886,4 +1886,11 @@ def build_http():
         http_timeout = socket.getdefaulttimeout()
     else:
         http_timeout = DEFAULT_HTTP_TIMEOUT_SEC
-    return httplib2.Http(timeout=http_timeout)
+    http = httplib2.Http(timeout=http_timeout)
+    # 308's are used by several Google APIs (Drive, YouTube)
+    # for Resumable Uploads rather than Permanent Redirects.
+    # This asks httplib2 to exclude 308s from the status codes
+    # it treats as redirects
+    http.redirect_codes = frozenset(set(http.redirect_codes) - {308})
```
plamut:
```suggestion
    http.redirect_codes = http.redirect_codes - {308}
```

(nit)

Actually, if `http.redirect_codes` is already a `frozenset`, we can remove 308 directly, and the result will already be a frozenset. Sorry for the misleading wording in the initial comment :/

```py
>>> fs = frozenset([1, 2, 3])
>>> fs
frozenset({1, 2, 3})
>> fs - {2}
frozenset({1, 3})
```

busunkim96: Thank you for taking the time to explain this! 😄

PR #823, opened 2020-02-12
docs: Update "Google Apps" references and add DwD example

eesheesh:


busunkim96:
Thank you for the update @eesheesh!

PR #824, opened 2020-02-14
MDM General

rpm-1973:


busunkim96:
Hi @rpm-1973! We already have a process for publishing to PyPI. Is there another issue you were hoping to address with this PR?

busunkim96:
Hi @rpm-1973,

We currently use an internal CI system called Kokoro (see .kokoro) directory for this library.

PR #841, opened 2020-03-12
fix: don't set http.redirect_codes if the attr doesn't exist

busunkim96:
This allows folks who can't upgrade their httplib2 version to use the latest version of the library. Note that it is still strongly recommended that you upgrade (see [install_requires](https://github.com/googleapis/google-api-python-client/blob/master/setup.py#L36)). 

@aaltay is there an issue tracking the httplib2 upgrade I can link to? I'd like this to be removed eventually.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ from setuptools import setup
 packages = ["apiclient", "googleapiclient", "googleapiclient/discovery_cache"]
 
 install_requires = [
-    "httplib2>=0.17.0,<1dev",
+    "httplib2>=0.9.2,<1dev",
     "google-auth>=1.4.1",
     "google-auth-httplib2>=0.0.3",
     "google-api-core>=1.13.0,<2dev",
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ from setuptools import setup
 packages = ["apiclient", "googleapiclient", "googleapiclient/discovery_cache"]
 
 install_requires = [
-    # NOTE: Apache Beam tests depend on this library and cannot currently
+    # NOTE: Apache Beam tests depend on this library and cannot
     # currently upgrade their httplib2 version.
     # Please see https://github.com/googleapis/google-api-python-client/pull/841
     "httplib2>=0.9.2,<1dev",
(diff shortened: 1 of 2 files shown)


aaltay:
Could we also change install requires to be similar to before like: `"httplib2>=0.9.2,<1dev"`. Because, now with this change the library will be backward compatible, and it will not cause force upgrades of httplib2 library. (force upgrades also cause issues.) ?

busunkim96:
Made the change to setup.py.

Is it possible for your tests to pin to an older version of this library until you're able to sort out upgrading httplib2? Or is there a requirement for you to always be testing with the latest version of google-api-python-client?

aaltay:
> Made the change to setup.py.
> 
> Is it possible for your tests to pin to an older version of this library until you're able to sort out upgrading httplib2? Or is there a requirement for you to always be testing with the latest version of google-api-python-client?

I don't there is a requirement to use at head. This is our setup file (https://github.com/apache/beam/blob/master/sdks/python/setup.py) -- I cannot find a direct dependency. Probably there is some indirect dependency.

setup.py
```diff
@@ -33,7 +33,7 @@
 packages = ["apiclient", "googleapiclient", "googleapiclient/discovery_cache"]
 
 install_requires = [
-    "httplib2>=0.17.0,<1dev",
+    "httplib2>=0.9.2,<1dev",
```
crwilcox: should we add a comment on this line so folks don't mistakenly try to update it?

busunkim96: Done.

setup.py
```diff
@@ -33,7 +33,10 @@
 packages = ["apiclient", "googleapiclient", "googleapiclient/discovery_cache"]
 
 install_requires = [
-    "httplib2>=0.17.0,<1dev",
+    # NOTE: Apache Beam tests depend on this library and cannot currently
+    # currently upgrade their httplib2 version.
```
daikeshi: nit: two "currently" here.

aaltay:
Thank you all! When would this change be available in pypi?

busunkim96:
Just looking for an approval on the release PR(#842) and then I'll push it out to PyPI. :)

busunkim96:
1.8.0 is released @aaltay https://pypi.org/project/google-api-python-client/1.8.0/

aaltay:
> 1.8.0 is released @aaltay https://pypi.org/project/google-api-python-client/1.8.0/

Thank you very much for the quick response!

PR #852, opened 2020-03-25
chore: use nox and release-please; drop 3.4

busunkim96:
* Replaces tox with nox
* Allows release-please releases

I attempted to make as few changes as possible with the tox -> nox transition. The lint rules are the existing standards and are different from other libraries.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -11,23 +11,34 @@ Thanks for stopping by to let us know something could be better!
 Please run down the following list and make sure you've tried the usual "quick fixes":
 
   - Search the issues already opened: https://github.com/googleapis/google-api-python-client/issues
-  - If you have a question, post on Stackoverflow under the `google-api` tag.
-  - If you are reporting an issue or requesting a feature for a G Suite API, please use their [public issue tracker](https://gsuite-developers.googleblog.com/2017/03/a-new-issue-tracker-for-g-suite.html)
+  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-python
```
plamut: Is this still relevant? Or should we be more specific and point the users to the main repository's README, as the latter contains the links to new split repos that are more likely to be monitored by the corresponding library's maintainers?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -11,23 +11,34 @@ Thanks for stopping by to let us know something could be better!
...
+  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+python
```
plamut:
```suggestion
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python
```
(nit) Let's use HTTPS by default to avoid and extra redirect to it (requesting an HTTP version results in 307). :)

.kokoro/build.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/bash
+# Copyright 2018 Google LLC
```
plamut: Is 2018 correct? Is it because that's the creation/addition date of this file?

Also, should config files contain the license header, too?

synth.py
```diff
@@ -0,0 +1,28 @@
...
+
+import synthtool as s
+from synthtool import gcp
+
+common = gcp.CommonTemplates()
+
+# ----------------------------------------------------------------------------
+# Add templated files
+# ----------------------------------------------------------------------------
+templated_files = common.py_library(unit_cov_level=100, cov_level=100)
```
plamut: We recently ditched the magic threshold numbers for individual unit test runs in other libraries, although having it here at 100 should be fine.

Is there a chance that a future fix would have to branch based on the Python version, causing the individual coverage check to start failing?

README.md
```diff
@@ -48,7 +48,7 @@ virtualenv <your-env>
 
 ## Supported Python Versions
 
-Python 3.4, 3.5, 3.6 and 3.7 are fully supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions
+Python 3.5, 3.6, 3.7, and 3.8 are fully supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions
```
plamut:
```suggestion
Python 3.5, 3.6, 3.7, and 3.8 are fully supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions.
```
(nit) The period at the end is missing.

.kokoro/build.sh
busunkim96: 2018 is the date this file was originally created (this is a template file).

The Open Source folks seem to recommend license headers for config files as well. Those are all changes in the synthtool templates. I'll open a PR there.

> any files that can be added to (i.e. anything that takes the format of a source file and supports file comments). This would include things like YAML files.

.github/ISSUE_TEMPLATE/bug_report.md
busunkim96: Ack, will tweak the template

.github/ISSUE_TEMPLATE/bug_report.md
busunkim96: Will tweak the template to be more specific when possible.

.kokoro/build.sh
busunkim96: https://github.com/googleapis/synthtool/pull/458

.github/ISSUE_TEMPLATE/bug_report.md
busunkim96: I deleted the bit with the google-cloud-python repo from the template. If they've already found this template, they've found their way to the issue tracker :)

setup.py
```diff
@@ -44,19 +46,21 @@
...
-version = googleapiclient.__version__
+version = "1.8.0"
 
 setup(
     name="google-api-python-client",
     version=version,
     description="Google API Client Library for Python",
-    long_description=long_desc,
-    author="Google Inc.",
+    long_description=readme,
```
mik-laj:
```suggestion
    long_description=readme,
    long_description_content_type='text/markdown'
```

busunkim96: Thanks for the review! I'll get a release out as soon as this PR is through @mik-laj

synth.py
busunkim96: I removed the threshold values as it wasn't being used. (The noxfile for this repo is handwritten). The existing coverage threshold for this library is [85%](https://github.com/googleapis/google-api-python-client/blob/master/tox.ini#L24) so I used that value in unit test session for `noxfile.py`.

setup.py
```diff
@@ -44,19 +46,22 @@
...
+    long_description_content_type='text/markdown'
```
mik-laj, line 62:
```suggestion
    long_description_content_type='text/markdown',
```

mik-laj: It breaks the builds. How else can I help in this change?

crwilcox:
Currently 3.4 tests are failing. We could drop it at this point, or fix what appears to be an encoding issue with code coverage tooling.

PR #871, opened 2020-04-20
Remove apiclient.__version__

### LINKED ISSUE #870: AttributeError: module 'googleapiclient' has no attribute '__version__'
When importing new version of google-api-python-client
`from apiclient import discovery`
i'm getting the error
`AttributeError: module 'googleapiclient' has no attribute '__version__'`

https://github.com/googleapis/google-api-python-client/blob/84d45619d753cb04d957651886231034194058b6/apiclient/__init__.py#L22

i guess this happens since you have removed `__version__` var from `googleapiclient/__init__.py`

https://github.com/googleapis/google-api-python-client/commit/f706cfd821ab7457e5db37abfc3619772657dd0e#diff-b926d296d4c856bcbf877809e4523562L15

can you please fix? @busunkim96 @mik-laj @crwilcox 

Traceback:
```    from apiclient import discovery
  File "/usr/local/lib/python3.7/site-packages/apiclient/__init__.py", line 22, in <module>
    __version__ = googleapiclient.__version__
AttributeError: module 'googleapiclient' has no attribute '__version__'

### PR DESCRIPTION
**Pull request by mik-laj**
Hello,

We have removed __version__ var from googleapiclient/__init__.py
https://github.com/googleapis/google-api-python-client/commit/f706cfd821ab7457e5db37abfc3619772657dd0e#diff-b926d296d4c856bcbf877809e4523562L15
so it won't work anymore.

https://github.com/googleapis/google-api-python-client/issues/870

CC: @busunkim96
Fixes #870 🐈

### DIFF
--- a/apiclient/__init__.py
+++ b/apiclient/__init__.py
@@ -19,8 +19,6 @@ except ImportError:
     sample_tools = None
 from googleapiclient import schema
 
-__version__ = googleapiclient.__version__
-
 _SUBMODULES = {
     "channel": channel,
     "discovery": discovery,

### CODE REVIEW

File: apiclient/__init__.py [hunk 1]
```diff
@@ -19,8 +19,6 @@
     sample_tools = None
 from googleapiclient import schema
 
-__version__ = googleapiclient.__version__
```
busunkim96, line 22: Would you mind getting the version from `pkg_resources` (in case someone is using this version for something)?


```
import pkg_resources

pkg_resources.get_distribution("google-api-python-client").version
```

mik-laj: Should I also restore this variable in `googleapiclient.__init__`?

busunkim96: I looked through this once more the only place __version__ was used is `model.py`, so I'm fine with deleting it. Thanks for opening this PR!

Issue #781, opened 2019-11-11
Reduce log spam by discovery.py by lowering logger level from info to debug/warn


timdonovanuk:
[This line](https://github.com/googleapis/google-api-python-client/blob/66bb32cc70353296dbf1876ae8c22baf6562cfb9/googleapiclient/discovery.py#L275) in discovery.py:

`logger.info("URL being requested: GET %s", actual_url)`

creates **a lot** of unncessary log spam. I don't think we need to know every time this code makes a request to Google, unless you're debugging. I use this library as part of HomeAssistant and my logs end up looking like (URLs sanitised):

```
2019-11-11 09:44:02 INFO (SyncWorker_6) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest
2019-11-11 09:44:02 INFO (SyncWorker_6) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/calendar/v3/calendars/
2019-11-11 09:44:02 INFO (SyncWorker_2) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest
2019-11-11 09:44:02 INFO (SyncWorker_2) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/calendar/v3/calendars/
2019-11-11 09:44:03 INFO (SyncWorker_9) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest
2019-11-11 09:44:03 INFO (SyncWorker_9) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/calendar/v3/calendars/
2019-11-11 09:44:03 INFO (SyncWorker_6) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest
2019-11-11 09:44:03 INFO (SyncWorker_6) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/calendar/v3/calendars/
2019-11-11 09:44:03 INFO (SyncWorker_6) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest
2019-11-11 09:44:04 INFO (SyncWorker_6) [googleapiclient.discovery] URL being requested: GET https://www.googleapis.com/calendar/v3/calendars/
[...]

PR #885, opened 2020-04-24
fix: downgrade repetitive logging calls to debug

busunkim96:
Reopens #782 from @bsolomon1124 as that branch is now gone.

The current INFO calls, which are highly repetitive, don't meet
the typical 'severity level' of an INFO message, and are more suited
for DEBUG.

Refs Issue #781


busunkim96:
CLA was signed in #782

PR #892, opened 2020-05-07
docs: changed http to https in setup, updated github links

mjip:


busunkim96:
Thank you for the patch @mjip!

PR #894, opened 2020-05-07
feat: allow to use 'six.moves.collections_abc.Mapping' in 'client_options.from_dict()'

paul1319:


paul1319:
@googlebot I fixed it.

PR #929, opened 2020-06-01
test: do not require unittest2

pgajdos:
I think unittest2 is no longer needed.


busunkim96:
@pgajdos Thank you for the patch! Would you be able to sign the CLA?

pgajdos:
Our layers would have to look at it and this would not be probably soon. So please proceed without my interaction, I will close this pull request afterwards.

PR #938, opened 2020-06-08
fix: prefer V2 Discovery URI format

busunkim96:
v2 is now preferred

229/263 APIs list the v2 style in the discovery directory.

https://www.googleapis.com/discovery/v1/apis

(Unit tests need to be updated to expect the v2 URI).

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -238,7 +238,7 @@ def build(
     else:
         discovery_http = http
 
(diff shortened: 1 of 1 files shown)


jay0lee:
I still see 29 instances of "https://www.googleapis.com/discovery/v1" in the doc and the v2 URL returns a hard 403 not found. I don't think we should make this change until all APIs support v2 URLs or at the very least we should try v2 and fallback to v1 if we get a 403.

PR #940, opened 2020-06-10
fix: update GOOGLE_API_USE_MTLS values

arithmetic1728:
Go client used lower case, so change the values to lower case in python apiary client.


arithmetic1728:
I think I will just use the lower case to be consistent with the Go implementation: https://github.com/googleapis/google-api-go-client/blob/master/transport/http/dial.go#L31-L35

PR #948, opened 2020-06-25
feat: Build universal wheels

nsoranzo:
See https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels


nsoranzo:
@googlebot I signed it!

PR #951, opened 2020-06-25
Added gdrive_api_tutorial.ipynb tutorial for beginners

mhdadk:
This tutorial demonstrates the many capabilities of the Google Drive Python API v3, including creating folders, uploading files, creating custom file properties, and searching using these file properties. I just started using the Google Drive Python API last week, and have been documenting my steps with it. Since this tutorial also contains many references, I feel that it will help new users (like myself) quickly get started with the API. There are a lot of challenges that I faced while using the API that I hope new users won't have to if they use this tutorial.

### DIFF
--- a/docs/gdrive_api_tutorial.ipynb
+++ b/docs/gdrive_api_tutorial.ipynb
@@ -1,0 +1,770 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Google Drive Python API Tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


mhdadk:
@googlebot I signed it!

mhdadk:
Hey @busunkim96 thanks for the help on the earlier issue. Any chance you can review this?

busunkim96:
Hi @mhdadk thank you for your interest in contributing! Modern samples for the Google Drive API generally live in https://github.com/gsuitedevs/python-samples. Please open a PR there for discussion.

mhdadk:
Hi @busunkim96, as you can see [here](https://github.com/gsuitedevs/python-samples/pull/165), I have opened a pull request in the repository that you suggested. However, nobody has commented on it and based on the history of pull requests in that repository, it is highly unlikely that someone will in the near future. Additionally, I only just found out about that repository when you mentioned it, which means that it is unlikely that another beginner such as myself would also know about this repository and therefore not be able to find my tutorial. Because of this, I think I should re-open this pull request. I really think this tutorial is most accessible to beginners if it is located in this repository.

busunkim96:
Hi @mhdadk,

The GSuite samples are maintained by a different team, so they should go in that repo.  We have a lot of samples spread out across APIs, so this is important for organization. In addition, we generally don't use notebooks for samples because they are hard to test. We require tests for all new samples because it lets us ensure samples are still working as intended.

If you're having trouble getting someone to look at your PR, it may be more appropriate to post this to another place (a blog perhaps?). Thank you again for your interest in helping other folks out!

mhdadk:
Hi @busunkim96, thanks a lot for the explanation. I completely understand.

PR #961, opened 2020-07-07
chore(deps)!: require 3.5+

busunkim96:
Removes 2.7 from the test suite and explicitly requires >= 3.5. 


parthea (commented):
2 minor observations, otherwise looks good!

ManikantSingh2710 (approved):

busunkim96:
@parthea Comments have been addressed, PTAL. :)

README.md
```diff
@@ -48,11 +48,11 @@ virtualenv <your-env>
 
 ## Supported Python Versions
 
-Python 3.5, 3.6 and 3.7, and 3.8 are fully supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions
+Python 3.6 and 3.7, 3.8, and 3.9 are fully supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions
```
parthea, line 51:
```suggestion
Python 3.6, 3.7, 3.8, and 3.9 are fully supported and tested. This library may work on later versions of 3, but we do not currently run tests against those versions.
```

parthea (approved):
LGTM, just a minor nit comment

PR #1030, opened 2020-09-14
fix(deps): require six>=1.13.0

### LINKED ISSUE #1029: googleapiclient.discovery.build fails with module 'six.moves' has no attribute 'collections_abc' since version 1.12.0
#### Environment details

  - OS type and version: 18.04.1-Ubuntu
  - Python version: Python 3.6.9
  - pip version: `pip --version` pip 9.0.1
  - `google-api-python-client` version: `pip show google-api-python-client`: Version: 1.12.0

#### Code example
googleapiclient.discovery.build() fails with message: module 'six.moves' has no attribute 'collections_abc'

We only see this problem with google-api-python-client 1.12.0. 1.11.0 is fine.

busunkim96:
This is needed for collections_abc

Fixes #1029  🦕

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ install_requires = [
     "google-auth>=1.16.0",
     "google-auth-httplib2>=0.0.3",
     "google-api-core>=1.21.0,<2dev",
-    "six>=1.6.1,<2dev",
+    "six>=1.13.0,<2dev",
     "uritemplate>=3.0.0,<4dev",
 ]
 


arithmetic1728 (approved):

celsooliveiraTBDE:
https://stackoverflow.com/questions/63902467/module-six-moves-has-no-attribute-collections-abc

I believe this may have broken something and the integrations aren't working as intended.

Issue #1047, opened 2020-09-26
'Http' object has no attribute 'close'


frankShih:
#### Environment details

  - OS type and version: ubuntu  16.04.6 LTS (4.15.0-1050-gcp)
  - Python version: 3.6.8
  - pip version: 3.6
  - `google-api-python-client` version:  1.7.1

  - Other dependency
![image](https://user-images.githubusercontent.com/8334979/94328892-e2c0c080-ffe8-11ea-936e-dffed74a1cf9.png)

```python
from apiclient.discovery import build
SCOPES = ['https://www.googleapis.com/auth/analytics.readonly']
KEY_FILE_LOCATION = conf.KEY_FILE_LOCATION

credentials = ServiceAccountCredentials.from_json_keyfile_name(
        KEY_FILE_LOCATION, SCOPES)

# Build the service object.
if version == 'v4':
    analytics = build('analyticsreporting', 'v4', credentials=credentials, cache_discovery=False)
elif version == 'v3':
    analytics = build('analytics', 'v3', credentials=credentials, cache_discovery=False)
return analytics
```

PS. `KEY_FILE_LOCATION` is a json key generated from GCP

#### Stack trace
```
  File "/home/dsu/ddt_dashboard/ddt_get_ga_report/lib/app.py", line 454, in get_report
    analyticsV3 = initialize_analyticsreporting('v3')
  File "/home/dsu/ddt_dashboard/ddt_get_ga_report/lib/app.py", line 117, in initialize_analyticsreporting
    analytics = build('analytics', 'v3', credentials=credentials, cache_discovery=False)
  File "/usr/local/lib/python3.6/dist-packages/googleapiclient/_helpers.py", line 134, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/googleapiclient/discovery.py", line 300, in build
    discovery_http.close()
AttributeError: 'Http' object has no attribute 'close'
```

The code works fine on my local machine (win10). 
Not sure if it is about package dependency or something else.

Any suggestions would be appreciated, thanks.

matabares:
I'm facing the same problem.

C:\Python363\python.exe N:/matabares/PythonSnippets/demoGoogleDrive.py
Traceback (most recent call last):
  File "N:/matabares/PythonSnippets/demoGoogleDrive.py", line 49, in <module>
    main()
  File "N:/matabares/PythonSnippets/demoGoogleDrive.py", line 34, in main
[...]

PR #1050, opened 2020-09-28
fix: Update setup.py to install httplib2>=0.15.0

tomchop:
I ran into the same issue described here: #1047

```
  File "/usr/local/lib/python3.6/dist-packages/googleapiclient/_helpers.py", line 134, in positional_wrapper
  File "/usr/local/lib/python3.6/dist-packages/googleapiclient/discovery.py", line 300, in build
    discovery_http.close()
AttributeError: 'Http' object has no attribute 'close'
```

From that same [issue](https://github.com/googleapis/google-api-python-client/issues/1047#issuecomment-699545589):

```
Having a look at httplib2 changelog, close() method appeared in v0.15.0
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,9 +35,6 @@ from setuptools import setup
 packages = ["apiclient", "googleapiclient", "googleapiclient/discovery_cache"]
 
 install_requires = [
-    # NOTE: Apache Beam tests depend on this library and cannot
-    # currently upgrade their httplib2 version.
-    # Please see https://github.com/googleapis/google-api-python-client/pull/841
     "httplib2>=0.15.0,<1dev",
     "google-auth>=1.16.0",
     "google-auth-httplib2>=0.0.3",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -38,7 +38,7 @@
     # NOTE: Apache Beam tests depend on this library and cannot
     # currently upgrade their httplib2 version.
     # Please see https://github.com/googleapis/google-api-python-client/pull/841
-    "httplib2>=0.9.2,<1dev",
+    "httplib2>=0.15.0,<1dev",
```
busunkim96, line 41: @aaltay Is it alright to bump this now?

aaltay: I _believe_ it can be bumped now. Beam has `'httplib2>=0.8,<0.18.0',` as the new range now. (https://github.com/apache/beam/blob/8bedfd402ccd1336f2b40fad58cd5833cecb9638/sdks/python/setup.py#L150)

Do you remember what was the specific problem?

(/cc @tvalentyn in case he remembers.)

tvalentyn: Yes, no need to restrict this for Beam reasons anymore.
https://issues.apache.org/jira/browse/BEAM-9819 has some context for prior restrictions.

busunkim96: Thanks!

tomchop: @busunkim96 Done!

busunkim96 (approved):
Thank you @tomchop!

PR #1088, opened 2020-11-12
docs: Update oauth docs to include snippet to get email address of authenticated user

### LINKED ISSUE #1071: Add user_info api in the documentation
I was able to retrieve user profile of the user using their credentials i.e. accessToken by the following code

```python

def get_user_email_from_credentials(credentials):
    user_info_service = build('oauth2', 'v2', credentials=credentials)
    user_info = user_info_service.userinfo().get().execute()
    return user_info['email']
```

But I couldn't find it in the docs here https://github.com/googleapis/google-api-python-client/blob/master/docs/dyn/index.md

parthea:
Fixes #1071  🦕

Note: I changed `scopes=['profile', 'email'],` to `scopes=['openid', 'https://www.googleapis.com/auth/userinfo.email', 'https://www.googleapis.com/auth/userinfo.profile'],` in the examples because I received a warning that the scope has changed.

> `Warning: Scope has changed from "email profile" to "https://www.googleapis.com/auth/userinfo.email openid https://www.googleapis.com/auth/userinfo.profile".`.


tswast (approved):
Thanks! I didn't know about this sample, but I have to do the same by trial-and-error when I want a set of scopes to debug auth logic.

I'm a bit disappointed that this is a code sample embedded in a markdown file (not testable), but that's a whole other issue.

busunkim96 (approved):

PR #1109, opened 2020-11-26
feat: Add support for using static discovery documents

### LINKED ISSUE #1063: Convert to a static/packaged library
The Python discovery client library `google-api-python-client` currently downloads the discovery doc for a service at runtime and constructs a client. Convert the library to a static/packaged library. See the [Node client](https://github.com/googleapis/google-api-nodejs-client/tree/master/src) for an example of this.

The Python client has a [discovery_cache](https://github.com/googleapis/google-api-python-client/tree/master/googleapiclient/discovery_cache) that is only usable with the deprecated auth client (oauth2client) or App Engine Standard's memcache.  One possible solution is adjusting this cache to reference downloaded json instead.

parthea:
- Add new parameter `static_discovery` to `googleapiclient.discovery.build()`. When `static_discovery` is set to `True` only static discovery documents included in the library will be used.
- If you prefer to not use static discovery documents, you can disable this feature by setting the `static_discovery` argument of `googleapiclient.discovery.build()` to `False`. For example :

```
build("youtubereporting", "v1", static_discovery=False)
```

Fixes #1063 🦕


bryanlaura736 (approved):

googleapiclient/discovery_cache/__init__.py
```diff
@@ -47,3 +48,36 @@ def autodetect():
...
+        A string containing the contents of the JSON discovery document,
+        otherwise None if the JSON discovery document was not found.
+    """
+
+    doc_name = None
+    content = None
+
+    # Extract the {apiVersion} and {api} from the uri which are the 2nd and 3rd
+    # last parts of the uri respectively.
+    # https://www.googleapis.com/discovery/v1/apis/{api}/{apiVersion}/rest
```
busunkim96, line 70: There are two different URI patterns (see `V1_DISCOVERY_URI` and `V2_DISCOVERY_URI` in `discovery.py`). Perhaps the service name and version can be passed from `build()` to `_retrieve_discovery_doc()` so the uri doesn't have to be parsed here?

googleapiclient/discovery_cache/__init__.py
```diff
@@ -47,3 +48,36 @@ def autodetect():
     except Exception as e:
         LOGGER.warning(e, exc_info=True)
         return None
+
+def get_static_doc(uri):
+    """Retrieves the discovery document from the directory defined in
+    DISCOVERY_DOC_STATIC_DIR corresponding to the uri provided.
```
busunkim96, line 54: Variable above is named `DISCOVERY_DOC_DIR`

googleapiclient/discovery_cache/__init__.py
parthea: Good catch! Fixed.

googleapiclient/discovery_cache/__init__.py
parthea: Good catch! Fixed.

bryanlaura736 (approved):

danoscarmike (approved):

busunkim96 (approved):

parthea:
@busunkim96 @SurferJeffAtGoogle I've updated the [docs](https://github.com/googleapis/google-api-python-client/pull/1109/files?short_path=298e509#diff-298e509e87a2e9501cb8f763fb811a38b5a10fb14ab30f4ef55ce7f0c58821b3), please can you take another look?

docs/start.md
```diff
@@ -65,6 +65,17 @@ with build('drive', 'v3') as service:
...
...h is only supported with `oauth2client<4.0`. If the artifact is not available in the legacy cache and the `static_discovery` argument of `build()` is set to `True`, the library will attempt to retieve the discovery artifact from the static copy of the discovery artifacts that are shipped with the library. As a last resort if the discovery artifact is not on disk, the discovery artifact will be retrieved from the internet. The maintainers recommend keeping `static_discovery=True`, which is the default, in order to avoid reliability issues related to transient network errors. The library currently does not detect breaking changes of discovery artifacts, and as a result, user code is not guaranteed to continue to succeed or fail with different versions of `google-api-python-client`. Users can set `static_discovery=False` as workaround if there is an incompatible discovery artifact shipped with the library. In addition, users may roll back the version of `google-api-python-client` if there is a breaking change in a discovery artifact. The maintainers of `google-api-python-client` encourage users to report issues with discovery artifacts so that the discovery artifacts can be corrected.
```
busunkim96, line 68: Thank you for adding this! It reads very nicely! 👍 

 One concern - `The library currently does not detect breaking changes of discovery artifacts` seems unsatisfying.  Users will have a hard time figuring out what changed between versions. NodeJS lists the added/deleted keys in the commit. https://github.com/googleapis/google-api-nodejs-client/commit/92b94ae7a3ace22248950961c6367eff2b3c15ff https://github.com/googleapis/google-api-nodejs-client/blob/master/CHANGELOG.md Can we do something similar in a follow up PR? @SurferJeffAtGoogle @JustinBeckwith

JustinBeckwith (approved):

bryanlaura736 (approved):

busunkim96 (approved):

docs/start.md
parthea: I agree. I've created #1154 to track this concern.

describe.py
```diff
@@ -395,19 +395,23 @@ def document_api(name, version, uri):
...
     try:
         service = build(name, version)
+
+        http = build_http()
+        response, content = http.request(
+            uri or uritemplate.expand(
+                FLAGS.discovery_uri_template, {"api": name, "apiVersion": version}
+            )
+        )
+        if 'status' in response:
```
SurferJeffAtGoogle, line 405: Why this change?

docs/start.md
```diff
@@ -65,6 +65,8 @@ with build('drive', 'v3') as service:
...
...t of `build()` is set to `True`, the library will attempt to retrieve the discovery artifact from the legacy cache which is only supported with `oauth2client<4.0`. If the artifact is not available in the legacy cache and the `static_discovery` argument of `build()` is set to `True`, the library will attempt to retieve the discovery artifact from the static copy of the discovery artifacts that are shipped with the library. The maintainers recommend keeping `static_discovery=True`, which is the default, in order to avoid reliability issues related to transient network errors. The library currently does not detect breaking changes of discovery artifacts, and as a result, user code is not guaranteed to continue to succeed or fail with different versions of `google-api-python-client`. Users can set `static_discovery=False` as workaround if there is an incompatible discovery artifact shipped with the library. In addition, users may roll back the version of `google-api-python-client` if there is a breaking change in a discovery artifact. The maintainers of `google-api-python-client` encourage users to report issues with discovery artifacts so that the discovery artifacts can be corrected.
```
SurferJeffAtGoogle, line 68: Could we describe this more in terms of the users' perspectives?  Something like the following

If reliability is most important, then set static_discovery=True.  It will use the service definition shipped in the library.
If always using the most recent version of the library is most important, then ...

SurferJeffAtGoogle (changes requested):

describe.py
parthea: Good catch! This change isn't necessary. `describe.py` is only used to generate the docs located [here](https://github.com/googleapis/google-api-python-client/tree/master/docs/dyn). After taking another look at function `document_api`, it should not be retrieving discovery artifacts over the internet so I removed that code. Please take another look.

docs/start.md
parthea: Done. I also removed the text about breaking change detection as that bug is captured in #1154 . Please can you take another look?

parthea:
@busunkim96 @SurferJeffAtGoogle Please can you take another look?

busunkim96 (approved):

docs/start.md
parthea: In an effort to move this PR forward, I'm going to resolve this comment as the feedback has been addressed.

describe.py
parthea: In an effort to move this PR forward, I'm going to resolve this comment as the feedback has been addressed.

PR #1120, opened 2020-12-04
docs: update guidance on service accounts

### LINKED ISSUE #946: Documentation does not explain how to create service account files
https://github.com/googleapis/google-api-python-client/blob/master/docs/oauth-server.md makes reference to `SERVICE_ACCOUNT_FILE` but doesn't explain where it comes from or what it contains.

Digging deeper leads me to https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.service_account.html#google.oauth2.service_account.Credentials which says:

"To create credentials using a Google service account private key JSON file"

but that doesn't explain how or where to get the file from either.

I did end up discovering https://stackoverflow.com/questions/46287267/how-can-i-get-the-file-service-account-json-for-google-translate-api which also confirms that if I haven't got the JSON file, I've got to create a new service account :(.

It would help if the documentation made the steps to setting up the service account clearer ...

busunkim96:
Link out to cloud.google.com docs on service accounts since it is less likely to go stale.

Fixes #946

### DIFF
--- a/docs/oauth-server.md
+++ b/docs/oauth-server.md
@@ -22,41 +22,14 @@ Finally, your application can use the access token to call Google APIs.
 
 ## Creating a service account
 
-A service account's credentials include a generated email address that is unique, a client ID, and at least one public/private key pair.
+https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating_service_account_keys
 
-If your application runs on Google App Engine, a service account is set up automatically when you create your project.
-
-If your application runs on Google Compute Engine, a service account is also set up automatically when you create your project, but you must specify the scopes that your application needs access to when you create a Google Compute Engine instance. For more information, see [Preparing an instance to use service accounts](https://cloud.google.com/compute/docs/authentication#using).
-
-If your application doesn't run on Google App Engine or Google Compute Engine, you must obtain these credentials in the Google API Console. To generate service-account credentials, or to view the public credentials that you've already generated, do the following:
-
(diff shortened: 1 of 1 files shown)


parthea (approved):
LGTM

javelinsubhash:
Not sure if this should be a new issue, but [docs](https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating_service_account_keys) show two different ways of creating keys with gcloud format 
```
{
  "type": "service_account",
  "project_id": "project-id",
  "private_key_id": "key-id",
  "private_key": "-----BEGIN PRIVATE KEY-----\nprivate-key\n-----END PRIVATE KEY-----\n",
  "client_email": "service-account-email",
  "client_id": "client-id",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://accounts.google.com/o/oauth2/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/service-account-email"
}
```
or REST API 
```
{
  "name": "projects/project-id/serviceAccounts/service-account-email/keys/key-id",
  "privateKeyType": "TYPE_GOOGLE_CREDENTIALS_FILE",
  "privateKeyData": "private-key",
  "validAfterTime": "date",
  "validBeforeTime": "date",
  "keyAlgorithm": "KEY_ALG_RSA_2048"
}
```

For this client library are both formats valid? I'm asking because I'm having auth issues with the gcloud format for this repository

PR #1126, opened 2020-12-06
fix: Improve support for error_details

### LINKED ISSUE #990: Not possible to provide `http` parameter in `next_chunk()` with resumable upload? HttpError 400
Hi,

This ticket is probably pointless, but maybe it hilights the need of a better error message?

The following code works fine with chunk size larger than 1024*1024, but using a chunksize of 1024 ends in a HttpError 400. Why is that so?

```python
from google.oauth2 import service_account
import googleapiclient.discovery as gad
import googleapiclient.http as gah

# Run once at start.
creds = service_account.Credentials.from_service_account_file(creds).with_scopes(['https://www.googleapis.com/auth/drive'])
drive = gad.build('drive', 'v3', credentials=creds)

# Run in a loop / implement both thread-safety and caching in http object.
filename = 'my_file.txt'
cache_path='.cache'
# Using 1024 ends up in an error.
# Using 1024*1024 works fine
# Using default parameter also works fine, but is actually not a resumable upload.
media = gah.MediaFileUpload(file_name, chunksize=1024, resumable=True) # Here
file_metadata = {'name': file_name, 'parents': [folder_id]}
request = self.drive.files().create(body=file_metadata, media_body=media, fields='id')
auth_http = _auth.authorized_http(creds)
auth_http.cache = httplib2.FileCache(cache_path)
response = None
while response is None:
        status, response = request.next_chunk(num_retries=4, http=auth_http)
```
Here is the error message I get when `chunksize=1024`.
```python
HttpError: <HttpError 400 when requesting https://www.googleapis.com/upload/drive/v3/files?fields=id&alt=json&uploadType=resumable returned "Bad Request">
```
Please, why is that so?

I thank you for your support.
Have a good day,
Bests,

parthea:

Fixes #990 🦕


googleapiclient/errors.py
```diff
@@ -48,7 +48,11 @@ def _get_reason(self):
         """Calculate the reason for the error from the response content."""
         reason = self.resp.reason
         try:
-            data = json.loads(self.content.decode("utf-8"))
+            try:
+                data = json.loads(self.content.decode("utf-8"))
+            except:
```
busunkim96, line 53: I'd prefer the specific exception be caught instead. It looks like it's `json.JSONDecodeError` in this case. https://docs.python.org/3/library/json.html#json.JSONDecodeError

parthea: Done

busunkim96 (approved):

PR #1147, opened 2020-12-23
fix: Catch ECONNRESET and other errors more reliably

### LINKED ISSUE #1146: http.py should catch more ECONNRESET conditions
Right now http.py implicitly assumes that an `ECONNRESET` error will be raised as a `ConnectionError`.  However, in some cases `urllib3` will raise this a regular OSError with the ernno set to match `ECONNRESET`.  http.py should check for the errnos that correspond to the subclasses of ConnectionError to be extra safe.

dossett:
Fixes #1146  🦕

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -28,9 +28,7 @@ __author__ = "jcgregorio@google.com (Joe Gregorio)"
 from six import BytesIO, StringIO
 from six.moves.urllib.parse import urlparse, urlunparse, quote, unquote
 
-import base64
 import copy
-import gzip
 import httplib2
 import json
 import logging
@@ -38,7 +36,6 @@ import mimetypes
 import os
 import random
 import socket
-import sys
 import time
 import uuid
 
@@ -190,11 +187,16 @@ def _retry_request(
             exception = connection_error
         except socket.error as socket_error:
             # errno's contents differ by platform, so we have to match by name.
+            # Some of these same errors may have been caught above, e.g. ECONNRESET *should* be
+            # raised as a ConnectionError, but some libraries will raise it as a socket.error
+            # with an errno corresponding to ECONNRESET
             if socket.errno.errorcode.get(socket_error.errno) not in {
                 "WSAETIMEDOUT",
                 "ETIMEDOUT",
                 "EPIPE",
                 "ECONNABORTED",
+                "ECONNREFUSED",
+                "ECONNRESET",
             }:
                 raise
             exception = socket_error
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -174,10 +174,10 @@ class HttpMockWithNonRetriableErrors(object):
             # set errno to a non-retriable value
             try:
                 # For Windows:
-                ex.errno = socket.errno.WSAECONNREFUSED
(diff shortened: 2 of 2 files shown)


dossett:
@googlebot I signed it!

monoueka:
Hi

monoueka:
SreyOn

googleapiclient/http.py
```diff
@@ -190,11 +187,16 @@ def _retry_request(
...
             # errno's contents differ by platform, so we have to match by name.
+            # Some of these same errors may have been caught above, e.g. ECONNRESET *should* be
+            # raised as a ConnectionError, but some libraries will raise it as a socket.error
+            # with an errno corresponding to ECONNRESET
             if socket.errno.errorcode.get(socket_error.errno) not in {
                 "WSAETIMEDOUT",
                 "ETIMEDOUT",
                 "EPIPE",
                 "ECONNABORTED",
+                "ECONNREFUSED",
```
parthea, line 198: @dossett Thanks for the fix! The test [TestHttpRequest.test_no_retry_connection_errors](https://github.com/googleapis/google-api-python-client/blob/master/tests/test_http.py#L937) is failing. I think it's because the test is [expecting OSError to be raised ](https://github.com/googleapis/google-api-python-client/blob/master/tests/test_http.py#L174) when the `socket.errno` is `ECONNREFUSED`. The test passes if I remove `ECONNREFUSED` from this list. Please can you take a look?

tests/test_http.py
```diff
@@ -174,10 +174,10 @@ def request(self, *args, **kwargs):
             # set errno to a non-retriable value
             try:
                 # For Windows:
-                ex.errno = socket.errno.WSAECONNREFUSED
+                ex.errno = socket.errno.WSAEHOSTUNREACH
```
dossett, line 177: Since `ECONNREFUSED` is now retried, we need to switch these to a different error that won't be retried

googleapiclient/http.py
dossett: Thanks for pointing that out @parthea.  I made a small change to `test_http` to have it use `EHOSTUNREACH` for errors that shouldn't be retried.  That's just the error that caught my eye when I looked through a list of error codes, but it does seem to work ok.

dossett:
I'll squash all the commits to conform to the commit message expectations when/if this is ok to merge.

.gitignore
```diff
@@ -6,6 +6,7 @@ dist/
 
 # Test files
 .nox/
+venv/
```
parthea, line 9: `venv` is likely the name of your specific virtual environment. Can we revert this change?

parthea (commented):
Thanks @dossett ! Just one minor observation.

.gitignore
dossett: whoops, thanks for spotting that!

monoueka:
Srey

parthea (commented):
Hi @dossett , The changes look good. Please could you update the commit message to have prefix `fix:`?

dossett:
@parthea commit message updated, thanks!

parthea (approved):
LGTM, Thanks @dossett !

PR #1161, opened 2021-01-13
fix: replace deprecated socket.error with OSError

### LINKED ISSUE #1148: http.py refers to socket.error, which is deprecated
It should just use `OSError` 
https://docs.python.org/3/library/socket.html#socket.error

dossett:
Fixes #1148 🦕


parthea (approved):
Thanks @dossett!

PR #1185, opened 2021-02-12
feat: add status_code property on http error handling

### LINKED ISSUE #1183: Get HTTP status code error handling
to improve error handling in the HttpError class, it would be useful to have a simpler way to get the request status code, as a `status code` property in the class.

eg.
```python
try:
    self.client.snapshots().get(
        project=[project name],
        snapshot=[description]
    ).execute()
except HttpError as err:
    if err.status_code == 404:
        do_something()
```

### LINKED ISSUE #1255: HttpError error_details isn't populated unless __repr__ is called first
I was trying to handle an HttpError by looking at the contents of the `error_details` attribute. I noticed the attribute is a null-string unless I trigger the `__repr__` function first. For example, this does not work as the error_details is always `""`. I made a simple test that demonstrates the error:

```
from googleapiclient import discovery, errors

client = discovery.build(
    "discovery", "v1"
)

req = client.apis().getRest(api='fake_api', version='v1')

try:
    resp = req.execute()
except errors.HttpError as err:
    print(f'Error details are currently: "{err.error_details}"')
    print(f'Exception string representation is: "{err}"')
    print(f'Error details are currently: "{err.error_details}"')
```

The output of the above code:

```
Error details are currently: ""
Exception string representation is: "<HttpError 404 when requesting https://www.googleapis.com/discovery/v1/apis/fake_api/v1/rest?alt=json returned "Requested entity was not found.". Details: "Requested entity was not found.">"
Error details are currently: "Requested entity was not found."
```

I tested and the behavior is the same on both `google-api-python-client-1.12.8` and `google-api-python-client-2.0.2`

wmarquardt:
Fixes #1183 🦕
Fixes #1255 🦕
- Add status code property in http error handler class.
- Resolve issue where error_details is not populated.


parthea (commented):
Hi @wmarquardt ,
Thanks for submitting a PR! Could you please also update the getting started documentation here with an example?

https://github.com/googleapis/google-api-python-client/blob/master/docs/start.md#execution-and-response

We could replace the existing example with 
```
try:
    response = request.execute()
except HttpError as e:
    print('Error response status code : {0}, reason : {1}'.format(e.status_code, e.error_details))
```

parthea:
Unrelated to your changes, when I ran the example code I found that e.error_details is empty in my test code. To fix it, we would need to make a change `errors.py`. The fix was to add a call to `self._get_reason()` in the `__init__` function [here](https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/errors.py#L45).

Change
```
    @util.positional(3)
    def __init__(self, resp, content, uri=None):
        self.resp = resp
        if not isinstance(content, bytes):
            raise TypeError("HTTP content should be bytes")
        self.content = content
        self.uri = uri
        self.error_details = ""
```
to
```
    @util.positional(3)
    def __init__(self, resp, content, uri=None):
        self.resp = resp
        if not isinstance(content, bytes):
            raise TypeError("HTTP content should be bytes")
        self.content = content
        self.uri = uri
        self.error_details = ""
        self._get_reason()
```

googleapiclient/errors.py
```diff
@@ -44,6 +44,10 @@ def __init__(self, resp, content, uri=None):
         self.uri = uri
         self.error_details = ""
 
+    @property
+    def status_code(self):
```
parthea, line 48:
```suggestion
    def status_code(self):
        """Return the HTTP status code from the response content."""
```

parthea (commented):
Hi @wmarquardt ,

I missed this in the last review. Please could you also add a docstring in the new function?

wmarquardt:
Hi @parthea,

Thank's for your review. I'll make these changes as soon as possible

busunkim96 (commented):
This looks OK to me @parthea but I am not very familiar with the code in `errors.py`. Is there something in particular you're concerned about?

busunkim96 (approved):

parthea (approved):

PR #1221, opened 2021-03-04
fix: Include discovery artifacts in published package

### LINKED ISSUE #1214: Published package is missing discovery files in discovery_cache
Many thanks to the @wyk9787 for noticing this and reaching out.

All calls to `discovery.build()` using `2.0.0` fail with "unknown api name or version".

```python
from googleapiclient import discovery

client = discovery.build("cloudprofiler", "v2")
```

This is because the published package has no `discovery_cache/documents` directory. 

1. `python3 -m venv env`
2. `source env/bin/activate`
3. `python3 -m pip install google-api-python-client`
4. `ls env/lib/python*/site-packages/googleapiclient/discovery_cache`

```
busunkim@busunkim:~/github$ ls env/lib/python*/site-packages/googleapiclient/discovery_cache
appengine_memcache.py  base.py  file_cache.py  __init__.py  __pycache__
```

parthea:
Fixes #1214  🦕

I was able to re-create the issue by using the following steps.
1) Create a clean `pyenv` environment
2) Download google-api-python-client-2.0.1 from PyPi [here](https://files.pythonhosted.org/packages/dc/67/a228c05a28288fdfde6e8c1a27a62e10ee265664c0f1d59070e6e0dc6242/google-api-python-client-2.0.1.tar.gz) and unzip it
3) Install it using `pip install <Unzipped Directory>`
4) Confirm `discovery.build()` fails on any call

To Verify the fix, repeat the steps above, but apply the update in this PR prior to installing the library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ setup(
     install_requires=install_requires,
     python_requires=">=3.6",
     packages=packages,
-    package_data={},
+    package_data={"googleapiclient": ["discovery_cache/documents/*.json"]},
     license="Apache 2.0",
     keywords="google api client",
     classifiers=[


busunkim96 (approved):

busunkim96:
Confirmed that the fix works for me locally. Thanks @parthea!

PR #1281, opened 2021-04-03
feat: Adds support for errors.py to also use 'errors' for error_details 

### LINKED ISSUE #1279: errors.py: Add support for JSON content that has `errors` instead of `detail` or `details`
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

 **Is your feature request related to a problem? Please describe.**

We have been using [errors.py](https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/errors.py) to raise HttpError errors. Unfortunately, many of our instances have content that matches the following format:
```
{
  error: {
    errors: {
    }
  }
}
```
versus what [_get_reason](https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/errors.py#L64) is looking for i.e. only one of `detail`, `details`, or `message`
This results us seeing the errors being logged but not the details:

```
<HttpError 503 when requesting url returned "Internal error. Please try again.">
```

 **Describe the solution you'd like**

The solution involves modifying [_get_reason](https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/errors.py#L64) to also support `error_detail_keyword` with an additional value of `errors`. This modifies the order to: `detail`, `details`, `errors`, `message`

 **Describe alternatives you've considered**
 **Additional context**

vinay-google:

Fixes #1279 🦕

### DIFF
--- a/googleapiclient/errors.py
+++ b/googleapiclient/errors.py
@@ -61,7 +61,7 @@ class HttpError(Error):
                 data = self.content.decode("utf-8")
             if isinstance(data, dict):
                 reason = data["error"]["message"]
-                error_detail_keyword = next((kw for kw in ["detail", "details", "message"] if kw in data["error"]), "")
+                error_detail_keyword = next((kw for kw in ["detail", "details", "errors", "message"] if kw in data["error"]), "")
                 if error_detail_keyword:
                     self.error_details = data["error"][error_detail_keyword]
             elif isinstance(data, list) and len(data) > 0:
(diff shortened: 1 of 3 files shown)


parthea (approved):
LGTM. Thanks for the fix @vinay-google!

PR #1292, opened 2021-04-14
chore: prevent normalization of semver versioning

dandhlee:
When there is a patch version added to semver versioning, setuptools.setup(version) will normalize the versioning from `-patch` to `.patch` which is not correct SEMVER versioning. The added feature with setuptools.sic(version) will prevent this from happening.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,22 @@ if sys.version_info < (3, 6):
 
 import io
 import os
-from setuptools import setup
+import setuptools
+
+# Disable version normalization performed by setuptools.setup()
+try:
+    # Try the approach of using sic(), added in setuptools 46.1.0
+    from setuptools import sic
+except ImportError:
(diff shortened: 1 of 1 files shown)


parthea:
This is on hold as @dandhlee confirmed that this requires a minimum version of `setuptools`.

dandhlee:
All is good now, if there is no problem with Kokoro pre-submits then it's fine, otherwise I'm working through the repositories to make necessary change and workarounds.

busunkim96:
@dandhlee This repo doesn't declare `setuptools` as a direct dependency, so please use the workaround to avoid breaking folks. 

You won't see a failure in CI since the tests will install the latest version of `setuptools` in a fresh environment.

dandhlee:
Understood, will make those changes around! Sorry for that 😓

parthea (approved):
LGTM

PR #1295, opened 2021-04-20
test: update nox file to build and test wheels

### LINKED ISSUE #1235: Modify noxfile to build and test the package
Versions `2.0.0` and `2.0.1` were yanked from PyPI last week due to an issue where discovery documents were not included in the published package causing `discovery.build()` to fail(#1214).  A basic check could be added to verify the package works correctly using the steps in #1214. Ideally it should be done on every PR and push to master so the issue can be caught before the package is published. 

Use these steps from #1214 to re-produce the issue with version `2.0.0` and `2.0.1`:
1. Start with a clean clone of `google-api-python-client`
2. Checkout version `2.0.0` or `2.0.1`, using `git checkout 2.0.0`
3. Run `python setup.py sdist`
4. Run `pip install dist/google-api-python-client-<version>.tar.gz`
5. Run 
```
$ python3
Python 3.8.7 (default, Jan 27 2021, 18:44:05) 
[GCC 10.2.1 20201224] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from googleapiclient import discovery
>>> client = discovery.build("cloudprofiler", "v2")
...
```

Before closing this issue, we should ensure that we have checks in place so that a PR will fail if `package_data` [here](https://github.com/googleapis/google-api-python-client/blob/master/setup.py#L78) is empty.

parthea:

Fixes #1235 🦕

In an attempt to verify that this is a fix for #1235  I'm going push a commit that removes the fix from #1221 in this PR and confirm that the checks fail.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -18,7 +18,6 @@ import sys
 import nox
 import os
 import shutil
-from shutil import rmtree
 
 test_dependencies = [
     "django>=2.0.0",
@@ -62,8 +61,8 @@ def lint(session):
 )
 def unit(session, oauth2client):
     # Clean up dist and build folders
-    rmtree('dist', ignore_errors=True)
-    rmtree('build', ignore_errors=True)
+    shutil.rmtree('dist', ignore_errors=True)
+    shutil.rmtree('build', ignore_errors=True)
 
     session.install(*test_dependencies)
     session.install(oauth2client)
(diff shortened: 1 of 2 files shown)


parthea:
I've confirmed that unit tests fail with error `../lib/python3.9/site-packages/googleapiclient/discovery.py:405: UnknownApiNameOrVersion` without the fix from #1221.

busunkim96 (approved):

Issue #1330, opened 2021-04-29
Duplicated bundled discovery documents


henribru:
It seems like some of the static discovery documents have ended up being duplicated under ever so slightly different names. One example is https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/admin.datatransfer_v1.json and https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/admin.datatransferv1.json. Another is https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/dfareporting.v3.3.json and https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/dfareporting.v33.json. There seems to be around 12 of these pairs. Just figured I'd let you know in case it was an accident.

parthea:
Thanks for reporting this, @henribru !

PR #1385, opened 2021-05-27
fix: resolve issue where certain artifacts would not be updated

### LINKED ISSUE #1384: Certain discovery artifacts are not being updated.
For example, `adexchangebuyer` version `v1.4` hasn't been updated in 6 months.

https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/adexchangebuyer.v1.4.json

This is related to #1330

parthea:
Fixes #1384

This PR is also a partial fix for #1330 where there are duplicate discovery artifacts for the same API and version. The issue is that in `describe.py` there is a call to `safe_version()` which is changing the version to remove decimal places.

For example, the following artifacts exist
https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/adexchangebuyer.v1.2.json
https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery_cache/documents/adexchangebuyer.v12.json

For `adexchangebuyer` version `v.1.2`, instead of  this
```
build('adexchangebuyer', 'v1.2')
```
we would have this.
```
build('adexchangebuyer', 'v12')
```

This PR fixes the versioning issue. The reason that this is a partial fix for #1330 is that another PR is needed to actually delete the duplicated discovery artifacts.

To validate this change, run [updatediscoveryartifacts.py](https://github.com/googleapis/google-api-python-client/blob/master/scripts/updatediscoveryartifacts.py) locally and you should see `adexchangebuyer.v1.2.json` updated.

Once this is merged, I'll trigger the GitHub action that updates discovery artifacts. This PR should not be a breaking change as it won't result in deleting discovery artifacts.

### DIFF
--- a/describe.py
+++ b/describe.py
@@ -403,9 +403,7 @@ def document_api(name, version, uri, doc_destination_dir):
     if resp.status == 200:
         discovery = json.loads(content)
         service = build_from_document(discovery)
-        version = safe_version(version)
-        doc_name = "{}.{}.json".format(name, version.replace("_", ""))
-
+        doc_name = "{}.{}.json".format(name, version)
         discovery_file_path = DISCOVERY_DOC_DIR / doc_name
         revision = None
 
@@ -435,7 +433,7 @@ def document_api(name, version, uri, doc_destination_dir):
         return
 
     document_collection_recursive(
-        service, "{}_{}.".format(name, version), discovery, discovery, doc_destination_dir
+        service, "{}_{}.".format(name, safe_version(version)), discovery, discovery, doc_destination_dir
     )
 
 


describe.py
```diff
@@ -435,7 +433,7 @@ def document_api(name, version, uri, doc_destination_dir):
         return
 
     document_collection_recursive(
-        service, "{}_{}.".format(name, version), discovery, discovery, doc_destination_dir
+        service, "{}_{}.".format(name, safe_version(version)), discovery, discovery, doc_destination_dir
     )
```
busunkim96, line 437: I think this change also needs to be made in `document_api_from_discovery_document`.

busunkim96 (changes requested):

describe.py
parthea: No change needed in `document_api_from_discovery_document`, as it's already being done here. https://github.com/googleapis/google-api-python-client/blob/master/describe.py#L457

parthea:
@busunkim96 PTAL

busunkim96 (approved):

PR #1404, opened 2021-06-07
fix: update content-length header for next page requests

### LINKED ISSUE #1403: _next method creates a HTTP request with wrong content-length header
Using google-api-python-client 2.2.0 and the GCP monitoring API, I received an error when retrieving the second page of a [projects.timeSeries.query](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/query) call:

```
can't fetch data, reason: Invalid JSON payload received. Unexpected end of string. Expected an object key or }.
```

Turns out that the content-length header is wrong and doesn't match with the body with the additional pageToken field.

The problem is in https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/discovery.py#L1285: the body is replaced, but the content-length stays the same because it is only updated at Request object creation time. I will prepare a PR for this.

schweikert:
The content-length header only gets updated at request object initialization, so make sure we update it here after we modified the body.

Fixes #1403

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -1283,6 +1283,9 @@ Returns:
             body = model.deserialize(request.body)
             body[pageTokenName] = nextPageToken
             request.body = model.serialize(body)
+            request.body_size = len(request.body)
+            if "content-length" in request.headers:
+              del request.headers["content-length"]
             logger.debug("Next page request body: %s %s" % (methodName, body))
 
         return request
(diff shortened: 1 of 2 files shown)


busunkim96 (commented):
Thank you! Could you add/modify a test that checks for this? https://github.com/googleapis/google-api-python-client/blob/4d1153db18a3edf86c5bb83149b4f1c0ba95f810/tests/test_discovery.py#L2136

schweikert:
> Thank you! Could you add/modify a test that checks for this?

Ah yes, thanks for spotting this! Done.

busunkim96 (approved):

PR #1512, opened 2021-09-01
chore: removing owlbot directives for conversion to main

nicain:
chore: removing owlbot directives for conversion to main


owlbot.py
```diff
@@ -41,6 +41,3 @@
 python.py_samples(skip_readmes=True)
 
 # Remove the replacements below once https://github.com/googleapis/synthtool/pull/1188 is merged
```
parthea, line 43:
```suggestion
```

parthea (commented):
Added one observation, otherwise LGTM.

PR #1553, opened 2021-10-06
feat: enable self signed jwt for service account credentials

arithmetic1728:
Enable self signed jwt if google-auth service account credentials are used.

Tested with Storage, Compute and PubSub APIs:

```
import googleapiclient.discovery

project = "<project>"
zone = "us-west1-a"

compute = googleapiclient.discovery.build('compute', 'v1')
result = compute.instances().list(project=project, zone=zone).execute()
print(result)

storage = googleapiclient.discovery.build('storage', 'v1')
result = storage.buckets().list(project=project).execute()
print(result)

topic = "<topic>"
pubsub = googleapiclient.discovery.build('pubsub', 'v1')
result = pubsub.projects().topics().get(topic=f"projects/{project}/topics/{topic}").execute()
print(result)
```

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -44,6 +44,7 @@ import uritemplate
 import google.api_core.client_options
 from google.auth.transport import mtls
 from google.auth.exceptions import MutualTLSChannelError
+from google.oauth2 import service_account
 
 try:
     import google_auth_httplib2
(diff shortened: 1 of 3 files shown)


busunkim96:
@arithmetic1728 I vaguely remember being told this was only implemented for GAPICs as self-signed JWT isn't guaranteed to work for all the APIs listed in the directory. Is that a non-issue now?

arithmetic1728:
@vriesk Thank you for bringing it up. I tested with the following sample, it did break. Will investigate more.

```
import googleapiclient.discovery

drive = googleapiclient.discovery.build('drive', 'v3', always_use_jwt_access=True)
result = drive.files().list(corpora="user").execute()
print(result)
```

Issue #1570, opened 2021-10-18
json.decoder.JSONDecodeError on csv (res) content


gparonitti:
#### Environment details

  - OS type and version: macOS 10.15.7 (19H1419)
  - Python version: 3.8.5
  - pipenv version: 2020.8.13
  - pip version: 20.2.3
  - `google-api-python-client` version: 2.26.1

#### Steps to reproduce

  Trying to use `generateCSV` api from `adsense.accounts.reports.saved.generateCsv`, but it might be a more general issue with non json format content.
  
#### Code example

```python
...
saved_report_id = 'accounts/pub-xxx/reports/yyy_zzz'

def main(argv):
  # Authenticate and construct service.
  credentials = adsense_util.get_adsense_credentials()
  with discovery.build('adsense', 'v2', credentials = credentials) as service:
    try:
      # Let the user pick account if more than one.
      account_id = adsense_util.get_account_id(service)

      # Retrieve report.
      # if saved_report_id:
      result = service.accounts().reports().saved().generateCsv(name=saved_report_id, dateRange='LAST_30_DAYS') .execute()

      print(result)
...
```

#### Stack trace
```
Traceback (most recent call last):
  File "src/generate_report.py", line 88, in <module>
    main(sys.argv)
  File "src/generate_report.py", line 63, in main
    print(result.execute())
  File "[...]/lib/python3.8/site-packages/googleapiclient/_helpers.py", line 131, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "[...]/lib/python3.8/site-packages/googleapiclient/http.py", line 939, in execute
    return self.postproc(resp, content)
  File "[...]/lib/python3.8/site-packages/googleapiclient/model.py", line 218, in response
    return self.deserialize(content)
  File "[...]/lib/python3.8/site-packages/googleapiclient/model.py", line 281, in deserialize
    body = json.loads(content)
  File "[...]/.pyenv/versions/3.8.5/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "[...]/.pyenv/versions/3.8.5/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/[...]/.pyenv/versions/3.8.5/lib/python3.8/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
[...]

PR #1574, opened 2021-10-19
fix bug #1570: manages JSONDecodeError exception

gparonitti:
test: copes with possible future regressions

### DIFF
--- a/googleapiclient/model.py
+++ b/googleapiclient/model.py
@@ -273,13 +273,18 @@ class JsonModel(BaseModel):
         return json.dumps(body_value)
 
     def deserialize(self, content):
+        body = {}
         try:
             content = content.decode("utf-8")
         except AttributeError:
             pass
-        body = json.loads(content)
-        if self._data_wrapper and isinstance(body, dict) and "data" in body:
-            body = body["data"]
+        try:
+            body = json.loads(content)
+            if self._data_wrapper and isinstance(body, dict) and "data" in body:
+                body = body["data"]
+        except json.decoder.JSONDecodeError:
+            if isinstance(content, str):
+                body = content
         return body
 
     @property
--- a/tests/test_json_model.py
+++ b/tests/test_json_model.py
@@ -31,7 +31,6 @@ import urllib
 
 import googleapiclient.model
 
-
 from googleapiclient.errors import HttpError
 from googleapiclient.model import JsonModel
 
@@ -289,6 +288,14 @@ class Model(unittest.TestCase):
         content = '{"data": "is good"}'
         content = model.response(resp, content)
         self.assertEqual(content, {"data": "is good"})
+
+    def test_no_data_wrapper_deserialize_text_format(self):
+        model = JsonModel(data_wrapper=False)
+        resp = httplib2.Response({"status": "200"})
+        resp.reason = "OK"
+        content = 'column1,column2,column3\nstring1,1.2,string2'
+        content = model.response(resp, content)
+        self.assertEqual(content, 'column1,column2,column3\nstring1,1.2,string2')
 
     def test_data_wrapper_deserialize(self):
         model = JsonModel(data_wrapper=True)


googleapiclient/model.py
```diff
@@ -273,13 +273,18 @@ def serialize(self, body_value):
...
-        body = json.loads(content)
-        if self._data_wrapper and isinstance(body, dict) and "data" in body:
-            body = body["data"]
+        try:
+            body = json.loads(content)
+            if self._data_wrapper and isinstance(body, dict) and "data" in body:
+                body = body["data"]
+        except json.decoder.JSONDecodeError:
+            if isinstance(content, str):
+                body = content
```
tseaver, line 287: Plese keep the scope of the `try:` minimal:
```suggestion
        try:
            body = json.loads(content)
        except json.decoder.JSONDecodeError:
            if isinstance(content, str):
                body = content
        else:
            if self._data_wrapper and isinstance(body, dict) and "data" in body:
                body = body["data"]
```

tests/test_json_model.py
```diff
@@ -290,6 +289,14 @@ def test_no_data_wrapper_deserialize(self):
         content = model.response(resp, content)
         self.assertEqual(content, {"data": "is good"})
 
+    def test_no_data_wrapper_deserialize_text_format(self):
+        model = JsonModel(data_wrapper=False)
+        resp = httplib2.Response({"status": "200"})
+        resp.reason = "OK"
+        content = 'column1,column2,column3\nstring1,1.2,string2'
```
tseaver, line 296: Please make this a named constant, and use it in the assertion below.

gparonitti: OK

googleapiclient/model.py
gparonitti: OK!

googleapiclient/model.py
```diff
@@ -273,13 +273,19 @@ def serialize(self, body_value):
         return json.dumps(body_value)
 
     def deserialize(self, content):
+        body = {}
```
tseaver, line 276: Shouldn't be needed (see below).

googleapiclient/model.py
```diff
@@ -273,13 +273,19 @@ def serialize(self, body_value):
...
         except AttributeError:
             pass
-        body = json.loads(content)
-        if self._data_wrapper and isinstance(body, dict) and "data" in body:
-            body = body["data"]
+        try:
+            body = json.loads(content)
+        except json.decoder.JSONDecodeError:
+            if isinstance(content, str):
+                body = content
```
tseaver, line 285: If the user passed in something other than `str`, we should just return it:
```suggestion
            body = content
```

googleapiclient/model.py
```diff
@@ -273,13 +273,19 @@ def serialize(self, body_value):
...
+        else:
+            if self._data_wrapper and isinstance(body, dict) and "data" in body:
+                body = body["data"]
```
tseaver, line 288: We can rely on `json.loads(content)` doing the Right Thing here, soe we dont need the type check:
```suggestion
        else:
            if self._data_wrapper and "data" in body:
                body = body["data"]
```

tests/test_json_model.py
```diff
@@ -290,6 +290,14 @@ def test_no_data_wrapper_deserialize(self):
...
         self.assertEqual(content, {"data": "is good"})
 
+    def test_no_data_wrapper_deserialize_text_format(self):
+        model = JsonModel(data_wrapper=False)
+        resp = httplib2.Response({"status": "200"})
+        resp.reason = "OK"
+        content = CSV_TEXT_MOCK
+        content = model.response(resp, content)
+        self.assertEqual(content, CSV_TEXT_MOCK)
+
```
tseaver, line 300: We need test coverage for the "oddball" cases:
- Users passes a non-`str` as `content`.
- `content` is valid JSON, but doesn't have a `data` key
- `content` has the expected `data` key, but the model has `data_wrapper` set to `False`.

tseaver (changes requested):

googleapiclient/model.py
gparonitti: That was to cope with "Local variable 'body' might be referenced before assignment" warning of my IDE, but I agree with removing it.

googleapiclient/model.py
gparonitti: ok

tests/test_json_model.py
gparonitti: - Ok for the first case.
- Doesn't the `test_data_wrapper_deserialize_nodata` test (the last one) address the second case?
- Ok for the third, but it feels very "oddball" :)

gparonitti: Ah, the third case seems to be addressed with `test_no_data_wrapper_deserialize` test.

gparonitti:
@googlebot I signed it!

tests/test_json_model.py
```diff
@@ -290,6 +291,23 @@ def test_no_data_wrapper_deserialize(self):
...
+    def test_no_data_wrapper_deserialize_raise_type_error(self):
+        buffer = io.StringIO()
+        buffer.write('String buffer')
+        model = JsonModel(data_wrapper=False)
+        resp = httplib2.Response({"status": "500"})
+        resp.reason = "The JSON object must be str, bytes or bytearray, not StringIO"
+        content = buffer
+        self.assertRaises(TypeError, model.response, (resp, content))
```
tseaver, line 309: We always prefer to use `assertRaises` as a context manager:
```suggestion
        with self.assertRaises(TypeError):
            model.response(resp, content)
```

tseaver (changes requested):
@gparonitti Thanks for persevering!  Just one more change needed.

PR #1616, opened 2021-12-02
fix: remove adsense.v1.4.json

parthea:
It is recommended to use adsense v2. See related bug b/202250173


parthea:
Wait until the release freeze ends on January 4, 2022.

busunkim96 (approved):

PR #1657, opened 2022-01-13
chore(v1): update keystore keyname

parthea:


.kokoro/release.sh
```diff
@@ -26,7 +26,7 @@ python3 -m pip install --upgrade twine wheel setuptools
 export PYTHONUNBUFFERED=1
 
 # Move into the package, build the distribution and upload.
-TWINE_PASSWORD=$(cat "${KOKORO_KEYSTORE_DIR}/73713_google_cloud_pypi_password")
+TWINE_PASSWORD=$(cat "${KOKORO_KEYSTORE_DIR}/73713_google-cloud-pypi-token-keystore-1")
 cd github/google-api-python-client
 python3 setup.py sdist bdist_wheel
 twine upload --username gcloudpypi --password "${TWINE_PASSWORD}" dist/*
```
busunkim96, line 32: I believe this line also needs to be updated, see https://github.com/googleapis/synthtool/blob/f8077d237e0df2cb0066dfc6e09fc41e1c59646a/synthtool/gcp/templates/python_library/.kokoro/release.sh#L32

busunkim96 (approved):

engelke (approved):

.kokoro/release.sh
parthea: Great catch! Fixed.

PR #1699, opened 2022-02-18
chore: reduce commits in discovery document update PR

parthea:
PR https://github.com/googleapis/google-api-python-client/pull/1698 contained more than 100 commits. Owl bot has a limit where it cannot support 100 commits in a PR.

The PR reduces the number of commits for the PR where discovery artifacts are updated. We can revert this change once https://github.com/googleapis/repo-automation-bots/issues/3203 is resolved.


busunkim96 (approved):

busunkim96:
Thanks for tracking down what happened with OwlBot!

PR #1707, opened 2022-02-28
Fix : Memory leak - issue while downloading large files #1706

### LINKED ISSUE #1706: Memory leak - issue while downloading large files
Issue :
next_chunk method stores downloaded content to 'fd' file object. This downloaded content is stored in ram. As the file size increases, it consumes more RAM to store downloaded data. 

Reference - https://github.com/googleapis/google-api-python-client/blob/5399bd3bf34ef31302a77ff484ef0067a18709a6/googleapiclient/http.py#L755

Related issues:
#512
#198

h3x4d1v1n3:
Implemented generator to keep RAM usage equal to the 'chunksize' over time. 

Fixes #1706

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -661,37 +661,32 @@ class MediaInMemoryUpload(MediaIoBaseUpload):
 
 
 class MediaIoBaseDownload(object):
-    """ "Download media resources.
+    """Download media resources.
 
     Note that the Python file object is compatible with io.Base and can be used
     with this class also.
 
-
     Example:
-      request = farms.animals().get_media(id='cow')
-      fh = io.FileIO('cow.png', mode='wb')
-      downloader = MediaIoBaseDownload(fh, request, chunksize=1024*1024)
+        request = farms.animals().get_media(id='cow')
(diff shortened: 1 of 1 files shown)


h3x4d1v1n3:
Hi team, what is the status of this pull request?

googleapiclient/http.py
```diff
@@ -661,37 +661,32 @@ def __init__(
...
     """
 
-    @util.positional(3)
-    def __init__(self, fd, request, chunksize=DEFAULT_CHUNK_SIZE):
+    @util.positional(2)
+    def __init__(self, request, chunksize=DEFAULT_CHUNK_SIZE):
         """Constructor.
 
         Args:
-          fd: io.Base or file object, The stream in which to write the downloaded
```
parthea, line 688: To avoid breaking existing users, ideally we should continue support the existing behaviour as well as the new desired behaviour.

parthea:
Hi @h3x4d1v1n3, I apologize for the delay in reviewing this PR. Please could you add a test and backwards compatible option?

h3x4d1v1n3:
For the backward compatibility I can add another function to the lattest version that will fix this issue.

PR #1782, opened 2022-05-04
chore: automatically format code using isort and black

parthea:
This PR adds a nox session to automatically sort python imports and runs black similar to the `format` session in the templated noxfile used in python-* googleapis repos [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2#L125).


noxfile.py
```diff
@@ -45,6 +59,26 @@ def lint(session):
...
+@nox.session(python="3.8")
+def format(session):
+    """
+    Run isort to sort imports. Then run black
+    to format code to uniform standard.
+    """
+    session.install(BLACK_VERSION, ISORT_VERSION)
+    # Use the --fss option to sort imports using strict alphabetical order.
+    # See https://pycqa.github.io/isort/docs/configuration/options.html#force-sort-within-sections
+    session.run(
```
m-strzelczyk, line 71: Aren't we usually using `--profile google` to sort? IIRC linter was complaining when I tried using isort with default profile.

PR #1792, opened 2022-05-06
chore: use isinstance() instead of type comparison

parthea:
Fixes warning from `flake8 . --select E721`

```
(py39) partheniou@partheniou-vm-2:~/git/google-api-python-client$ flake8 . --select E721
./googleapiclient/model.py:177:28: E721 do not compare types, use 'isinstance()'
./googleapiclient/model.py:396:37: E721 do not compare types, use 'isinstance()'
./googleapiclient/discovery.py:701:24: E721 do not compare types, use 'isinstance()'
./googleapiclient/discovery.py:712:24: E721 do not compare types, use 'isinstance()'
./googleapiclient/discovery.py:1078:66: E721 do not compare types, use 'isinstance()'
```

### DIFF
--- a/googleapiclient/discovery.py
+++ b/googleapiclient/discovery.py
@@ -698,10 +698,7 @@ def _cast(value, schema_type):
       A string representation of 'value' based on the schema_type.
     """
     if schema_type == "string":
-        if type(value) == type("") or type(value) == type(""):
-            return value
-        else:
-            return str(value)
+        return str(value)
     elif schema_type == "integer":
         return str(int(value))
     elif schema_type == "number":
@@ -709,10 +706,7 @@ def _cast(value, schema_type):
     elif schema_type == "boolean":
         return str(bool(value)).lower()
     else:
-        if type(value) == type("") or type(value) == type(""):
-            return value
-        else:
-            return str(value)
+        return str(value)
 
 
 def _media_size_to_long(maxSize):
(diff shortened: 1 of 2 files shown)


googleapiclient/discovery.py
```diff
@@ -698,21 +698,15 @@ def _cast(value, schema_type):
       A string representation of 'value' based on the schema_type.
     """
     if schema_type == "string":
-        if type(value) == type("") or type(value) == type(""):
```
atulep, line 701: This looks weird, but also potentially unsafe to remove.

Axel-Jacobsen: This is indeed weird, but is also safe.

Look at the last time it was changed (you will have to expand `discover.py` and scroll to line 701): https://github.com/googleapis/google-api-python-client/commit/3bbefc1352bcb2e302f7736643c9363799d5f5df#diff-b039ba2bb64e79cf4c77e4fe5a4b21a826172e8b7ae9be872fd2d46047c05cbbL695-R701

Originally, it was 

```python3
if type(value) == type("") or type(value) == type(u""):
```

But was changed automatically to be

```python3
if type(value) == type("") or type(value) == type(""):
```

Because `type("") == type(u"") == "str"`, this is certainly safe.

whoami-anoint (approved):

PR #1793, opened 2022-05-06
chore: fix boolean comparison

parthea:
Fixes warning from `flake8 . --select E712`

```
(py39) partheniou@partheniou-vm-2:~/git/google-api-python-client$ flake8 . --select E712
./scripts/buildprbody.py:83:40: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
./scripts/changesummary.py:211:46: E712 comparison to True should be 'if cond is True:' or 'if cond:'
./scripts/changesummary.py:223:50: E712 comparison to True should be 'if cond is True:' or 'if cond:'
```


vchudnov-g (approved):

msbatin (commented):
https://github.com/msbatin/google-api-python-client/pull/2#pullrequestreview-2799670616

daniel-sanche:
Closing due to inactivity

PR #1802, opened 2022-05-16
fix: Empty last chunk case in resumable upload that is not a stream

### LINKED ISSUE #1801: MediaUpload with has_stream() = False and an empty last chunk results in HTTP Error 400
#### Environment details

  - OS type and version: Ubuntu 20.04
  - Python version: Python 3.8.10
  - pip version: pip 20.0.2
  - `google-api-python-client` version: Version: 2.46.0

#### Steps to reproduce

  1. Create a class that inherit off MediaUpload, but resumable() = True and has_stream() = False so getbytes() will be used.
  2. Do an upload that is an exact multiples of the chunksize.
  3. The last chunk will be empty, and the header will look like this:
```
{'Content-Length': '0', 'Content-Range': 'bytes 268435456-268435455/268435456'}
```
  4. This will yield the HTTP 400 error code.

#### Code example

Note: Please fill in a folder ID in metadata['parents'] before running. 

```
from __future__ import print_function

import os.path

from google.auth.transport.requests import Request
from google.oauth2.credentials import Credentials
from google_auth_oauthlib.flow import InstalledAppFlow
from googleapiclient.errors import HttpError
from googleapiclient.http import MediaUpload

# If modifying these scopes, delete the file token.json.
SCOPES = ["https://www.googleapis.com/auth/drive", "https://www.googleapis.com/auth/drive.file", "https://www.googleapis.com/auth/drive.metadata.readonly"]

class ExampleUpload(MediaUpload):
  def __init__(self):
    self.data = b'\0'*(200*1024*1024)

  def chunksize(self):
    return 100*1024*1024

  def mimetype(self):
    return "application/octet-stream"

  def size(self):
    return None

  def resumable(self):
    return True

  def has_stream(self):
    return False

  def getbytes(self, begin, length):
    length = min(len(self.data)-begin, length)
    return self.data[begin:begin+length]

def main():
    """Shows basic usage of the Drive v3 API.
    Prints the names and ids of the first 10 files the user has access to.
    """
    creds = None
    # The file token.json stores the user's access and refresh tokens, and is
    # created automatically when the authorization flow completes for the first
    # time.
    if os.path.exists('token.json'):
        creds = Credentials.from_authorized_user_file('token.json', SCOPES)
    # If there are no (valid) credentials available, let the user log in.
    if not creds or not creds.valid:
        if creds and creds.expired and creds.refresh_token:
            creds.refresh(Request())
        else:
            flow = InstalledAppFlow.from_client_secrets_file(
                'credentials.json', SCOPES)
            creds = flow.run_local_server(port=0)
        # Save the credentials for the next run
        with open('token.json', 'w') as token:
            token.write(creds.to_json())

    try:
        service = build('drive', 'v3', credentials=creds)

        metadata = {}
        metadata['name'] = 'test_file1'
        metadata['parents'] = <FOLDER ID HERE>
        media_body = ExampleUpload()
        request = service.files().create(body=metadata, media_body=media_body)
        response = None
        while response is None:
          status, response = request.next_chunk()
          print('Got status and response: %s %s'%(status, response))
    except HttpError as error:
        # TODO(developer) - Handle errors from drive API.
        print(f'An error occurred: {error}')


if __name__ == '__main__':
    main()
```

#### Stack trace
```
  File "/venv/lib/python3.8/site-packages/googleapiclient/_helpers.py", line 131, in positional_wrapper
  File "/venv/lib/python3.8/site-packages/googleapiclient/http.py", line 1097, in next_chunk
    return self._process_response(resp, content)
  File "/venv/lib/python3.8/site-packages/googleapiclient/http.py", line 1128, in _process_response
    raise HttpError(resp, content, uri=self.uri)
googleapiclient.errors.HttpError: <HttpError 400 when requesting https://www.googleapis.com/upload/drive/v3/files?alt=json&uploadType=resumable returned "Bad Request". Details: "Failed to parse Content-Range header.">
```

john0312:
For an upload that inherits the MediaUpload class and is not a stream,
if the size of the upload is exact multiples of the chunksize, the last
chunk would be empty and the generated Content-Range header will result
in "Bad Request". Details: "Failed to parse Content-Range header."

Fixes #1801 🦕

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1062,10 +1062,14 @@ class HttpRequest(object):
             "Content-Length": str(chunk_end - self.resumable_progress + 1),
         }
 
-        # An empty file results in chunk_end = -1 and size = 0
-        # sending "bytes 0--1/0" results in an invalid request
-        # Only add header "Content-Range" if chunk_end != -1
+        if chunk_end - self.resumable_progress + 1 == 0 and chunk_end != -1:
+            # This is the last chunk, it's empty but the file is not empty.
+            # Send a Content-Range that ends this file.
+            headers["Content-Range"] = "bytes */%s" % (size,)
         if chunk_end != -1:
+            # An empty file results in chunk_end = -1 and size = 0
+            # sending "bytes 0--1/0" results in an invalid request
+            # Only add header "Content-Range" if chunk_end != -1
             headers["Content-Range"] = "bytes %d-%d/%s" % (
                 self.resumable_progress,
                 chunk_end,


Blessedlybroken:
> For an upload that inherits the MediaUpload class and is not a stream,
> 
> if the size of the upload is exact multiples of the chunksize, the last
> 
> chunk would be empty and the generated Content-Range header will result
> 
> in "Bad Request". Details: "Failed to parse Content-Range header."
> 
> 
> 
- [ ] > Fixes #1801 `🦕 
> 
>

Blessedlybroken (approved):

googleapiclient/http.py
```diff
@@ -1062,10 +1062,14 @@ def next_chunk(self, http=None, num_retries=0):
             "Content-Length": str(chunk_end - self.resumable_progress + 1),
         }
 
-        # An empty file results in chunk_end = -1 and size = 0
-        # sending "bytes 0--1/0" results in an invalid request
-        # Only add header "Content-Range" if chunk_end != -1
+        if chunk_end - self.resumable_progress + 1 == 0 and chunk_end != -1:
```
vchudnov-g, line 1065: This `if` only fires if `chunk_end != -1`. But in that case, the `if` below also fires, overwriting this particular header. So this block as written has no effect on the end result.

john0312: Sorry for the typo, resolved in the force push.

googleapiclient/http.py
```diff
@@ -1062,10 +1062,14 @@ def next_chunk(self, http=None, num_retries=0):
             "Content-Length": str(chunk_end - self.resumable_progress + 1),
         }
 
-        # An empty file results in chunk_end = -1 and size = 0
-        # sending "bytes 0--1/0" results in an invalid request
-        # Only add header "Content-Range" if chunk_end != -1
-        if chunk_end != -1:
+        if chunk_end - self.resumable_progress + 1 == 0 and chunk_end != -1:
```
vchudnov-g, line 1065: I suggest assigning 

```
content_length = chunk_end - self.resumable_progress + 1 == 0 and chunk_end
```

before the `headers = ...` line and using `content_length` both for the header and in this conditional.

vchudnov-g (approved):
PTAL at my suggestion for clarity/code-deduping.

Otherwise, LGTM

googleapiclient/http.py
chalmerlowe: @vchudnov-g 

I am not sure I am tracking here on two things. 

#### One: Do you mean for this to equate to a boolean?
This line from your example:
```
chunk_end - self.resumable_progress + 1 == 0 and chunk_end
```
yields `True` or `False`.

And yet `content_length` feels like the name for an numerial value.

#### Two: Can you simply suggest the code edit you want to see using GitHub's suggestion tool?

I am not tracking exactly what you want in terms of placement and modifications of the new code. Are you suggesting the following? This seems like doesn't do what we want (especially if `content_length` is a boolean)?
```suggestion
        content_length = chunk_end - self.resumable_progress + 1 == 0 and chunk_end != -1
        if content_length:
            headers["Content-Range"] = "bytes */%s" % (content_length,)
```

john0312: Agrees with @chalmerlowe and from my understanding of this change, I believe it might be hard to simplify the logic here. I propose we merge this as is, if there are further logic simplifications possible, I suggest we resolve through further PRs.

Note that this commit is authored while I'm still employed at Google so merging it as is would have much lesser copyright related hoops to jump through.

PR #1864, opened 2022-08-03
test: Switch to unittest.mock

### LINKED ISSUE #1863: Stop using external 'mock' dependency
As of Python 3.4, 'mock' is included in the standard library under the unittest module, and since the lowest supported version of Python is greater than that, we can remove the external dependency.

s-t-e-v-e-n-k:
As of Python 3.4, the unittest module in the standard library has
included mock, which means we can remove one external dependency.

Fixes #1863

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -42,7 +42,6 @@ test_dependencies = [
     "pytest-cov",
     "webtest",
     "coverage",
-    "mock",
 ]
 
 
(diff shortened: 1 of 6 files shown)


parthea (approved):

parthea:
Thanks @s-t-e-v-e-n-k !

PR #2021, opened 2023-01-05
test: Support running the test suite without oauth2client

mgorny:
Make it possible to run the test suite without oauth2client, by skipping the few tests that specifically require it.  Given that oauth2client is deprecated, we are aiming towards removing it entirely from Gentoo and the unconditional imports in the test suite are a blocker for that.

I've also added running without oauth2client as an additional nox scenarios but I can remove that if you prefer.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -84,6 +84,7 @@ def format(session):
 @nox.parametrize(
     "oauth2client",
     [
+        None,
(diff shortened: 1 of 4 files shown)


parthea (approved):
Thanks @mgorny!

mgorny:
Thanks a lot!

PR #2033, opened 2023-01-12
Update setup.py with Find package instead of doing it manually!

aria1991:
This enhanced script includes the following additional functionality:

— The `find_packages` method is used instead of manually specifying the packages. This ensures that all packages in the distribution are included, and excludes any test packages.

— Additional entry points for command line scripts are added. This allows users to run the `googleapiclient_cli` script directly from the command line, without having to specify the full path to the script.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,45 +20,47 @@ are not already installed.
 from __future__ import print_function
 
 import sys
+import io
+import os
+from setuptools import find_packages, setup
 
 if sys.version_info < (3, 7):
     print("google-api-python-client requires python3 version >= 3.7.", file=sys.stderr)
     sys.exit(1)
 
-import io
-import os
+# Additional package data to include in the distribution
+package_data = {
+    "googleapiclient": [
+        "discovery_cache/documents/*.json",
+        "discovery_cache/schemas/*.json"
+    ]
+}
 
-from setuptools import setup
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -87,4 +90,15 @@
...
     ],
+    # Additional functionality
+    # Scripts to include in the distribution
+    scripts=["scripts/googleapiclient_cli.py"],
+    # Additional data files to include in the distribution
+    data_files=[("config", ["config/googleapiclient.cfg"])],
+    # Additional entry points for command line scripts
+    entry_points={
+        "console_scripts": [
+            "googleapiclient_cli = googleapiclient.scripts.googleapiclient_cli:main"
```
parthea, line 101: Hi @aria1991,
The module `googleapiclient_cli` does not exist [here](https://github.com/googleapis/google-api-python-client/tree/main/scripts). Please could you take a look?

aria1991: Hi @parthea thanks for your comment. My bad! I added to the `keywords` in my new commit, you can ignore it. Do you have any suggestion on adding it (sth like requirement text file) otherwise we can ignore this part and remove it!

parthea:
Closing as obsolete. Please open an issue first to discuss this feature request.

PR #2099, opened 2023-04-01
fix: Removed indent from discovery cache json files

### LINKED ISSUE #1967: Consider minifying json shipped in package
tl;dr we are deploying the SDK to aws lambdas and ran into a scenario where our primary layer became too large, initially we just stripped individual json files from discovery_cache/documents since this is by far the largest culprit (according to `ncdu`).

I also noticed that all of these definitions are actually pretty printed json.

I'm suggesting 2 things:
1. Ship a utility to select (and delete other) json definitions that the user wants to use in a deployment.
2. Minify all json contents (see trivial script below), this reduced the package size (on disk) by approximately 20mb (and probably also improves load times although I did not test for that).

```js
fs.readdirSync('.').forEach(f => {
  const data = JSON.parse(fs.readFileSync(f, 'utf8'));
  fs.writeFileSync(f, JSON.stringify(data));
  console.log('processed', f);
});
```

#### Environment details

  - OS type and version: irrelevant
  - Python version: 3.9
  - pip version: 22.2.2
  - `google-api-python-client` version: 2.64.0

#### Steps to reproduce

  1. Install package
  2. 72mb via `ncdu`

#### Code example

n/a

#### Stack trace

n/a

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

jorenham:
This should bring down the size by 15-30%.
Towards #1967


jorenham:
I gotta say, the amount of effort that is required to propose this **1 character** change is borderline Kafkaesque.

parthea (approved):

PR #2125, opened 2023-05-08
build: add describe.py to scripts folder to fix failing Github Action

parthea:
[The Github Action that is used to update the cache of discovery artifacts](https://github.com/googleapis/google-api-python-client/blob/main/.github/workflows/main.yml) recently started failing with error `ModuleNotFoundError: No module named 'describe'` without any code changes in this repo. See build log [here](https://github.com/googleapis/google-api-python-client/actions/runs/4858525217/jobs/8660053081).

Through testing in https://github.com/googleapis/google-api-python-client/pull/2124, I found that moving the `describe.py` module to the `scripts` folder solved the issue. This module is only used by the GitHub action and is not part of `googleapiclient`.

Below is the specific line that was failing in the GitHub action:
https://github.com/googleapis/google-api-python-client/blob/84530bde51148a9c3e104f4e4622689e479bfed8/scripts/updatediscoveryartifacts.py#L23


noxfile.py
```diff
@@ -138,7 +137,7 @@ def scripts(session):
         "--cov=scripts",
         "--cov-config=.coveragerc",
         "--cov-report=",
-        "--cov-fail-under=91",
+        "--cov-fail-under=59",
```
vchudnov-g, line 140: Why are we decreasing the coverage here? I would keep it as high as we can, so we can be notified as early as possible of any breakages.

scripts/describe.py
vchudnov-g, line 1: This file is current a CLI tool at the root level that people may be looking for. Moving it arguably qualifies as a breaking change. How about we make one of the copies a symlink to the other?

vchudnov-g (changes requested):
Thank you for finding the workaround, but I am disturbed that it was working before and then suddenly stopped working. Did a path change? Or did a `pip install` like [this one](https://github.com/googleapis/google-api-python-client/blob/84530bde51148a9c3e104f4e4622689e479bfed8/.github/workflows/main.yml#L49) not run as it used to, or fail? It makes me uncomfortable to fix this blindly without knowing why it broke.

Aside from that, and another question about coverage, blocking on one change: symlink `describe.py` so we don't change the surface.

noxfile.py
parthea: Reverted

describe.py
```diff
@@ -356,7 +356,12 @@ def document_collection(resource, path, root_discovery, discovery, css=CSS):
 
 
 def document_collection_recursive(
-    resource, path, root_discovery, discovery, doc_destination_dir
+    resource,
+    path,
+    root_discovery,
+    discovery,
+    doc_destination_dir,
+    artifact_destination_dir=DISCOVERY_DOC_DIR,
```
vchudnov-g, line 364: Since this script is now accessible via the symlink from multiple directories, we have to ensure the `DISCOVER_DOC_DIR` default is set correctly above (line ~40). Shouldn't that line be `DISCOVERY_DOC_DIR = (pathlib.Path(__file__).resolve().parent / …)` rather than `...parent.resolve()…` as we have now?

IIUC, the way the variable is defined now makes sense for the script being at the repo top level, but if it's being called as a symlink in the `scripts/` subdirectory, it's not clear `__file__` will resolve to the link target, so it seems we need to `resolve()` as early as possible so we know where we are.

vchudnov-g (commented):
One question about `DISCOVERY_DOC_DIR` in the script. 

And still wondering what made this stop working all of a sudden.

parthea:
> And still wondering what made this stop working all of a sudden.

I was unable to reproduce the original issue outside of Github Actions. It seems to me that something changed in the Github Actions environment which caused this action to fail. The changes in this PR seem to resolve the issue.

vchudnov-g:
OK. We need to ensure there is (and we pay attention to) proactive communication about breakages like this. It's not good to be taken by surprise.

PR #2152, opened 2023-06-06
fix: remove faulty version matcher in setup.py

### LINKED ISSUE #2151: `<1dev` is not a valid matcher for Python package versions in `setup.py`!
On the current master branch, you have this matcher for an `install_requires` package version in [setup.py](https://github.com/googleapis/google-api-python-client/blob/110667251e8b2c8852945bfe238f399742148cda/setup.py#L36):

```python
install_requires = [
    "httplib2>=0.15.0,<1dev",
```

The `<1dev` part is not a valid in the sense of [PEP 440](https://peps.python.org/pep-0440/) and cause problems with a number of packing tools, as seen here:

```shell
$ python -c 'from distlib.version import NormalizedMatcher; NormalizedMatcher("httplib2>=0.15.0,<1dev")'
  File "<string>", line 1, in <module>
  File "/home/jan/.local/share/virtualenvs/master_clean-V3vlFZeD/lib/python3.7/site-packages/distlib/version.py", line 125, in __init__
    vn, prefix = self.version_class(s), False
  File "/home/jan/.local/share/virtualenvs/master_clean-V3vlFZeD/lib/python3.7/site-packages/distlib/version.py", line 33, in __init__
    self._parts = parts = self.parse(s)
  File "/home/jan/.local/share/virtualenvs/master_clean-V3vlFZeD/lib/python3.7/site-packages/distlib/version.py", line 267, in parse
    result = _normalized_key(s)
  File "/home/jan/.local/share/virtualenvs/master_clean-V3vlFZeD/lib/python3.7/site-packages/distlib/version.py", line 188, in _pep_440_key
    raise UnsupportedVersionError('Not a valid version: %s' % s)
distlib.version.UnsupportedVersionError: Not a valid version: 1dev
```

This is just speculation, but it looks like it was either faultily auto-generated by some tool, or is an incomplete copy & paste. This should be corrected and ideally the artifact files of the effected versions corrected on pypi.org.

con-f-use:
`httplib2<1dev` is ~not~ a valid version matcher (edit: it is valid, just doesn't parse with distlib) in the sense of [PEP 440](https://peps.python.org/pep-0440/).

> The canonical public version identifiers MUST comply with the following scheme:
> `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`

It is also considered bad practice to have an upper bound on package versions and installers like pip do not consider development versions in any case (unless explicitly told to).

Fixes #2151

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,16 +33,16 @@ from setuptools import setup
 packages = ["apiclient", "googleapiclient", "googleapiclient/discovery_cache"]
 
 install_requires = [
-    "httplib2>=0.15.0,<1dev",
+    "httplib2>=0.15.0,<1.dev0",
     # NOTE: Maintainers, please do not require google-auth>=2.x.x
     # Until this issue is closed
     # https://github.com/googleapis/google-cloud-python/issues/10566
-    "google-auth>=1.19.0,<3.0.0dev",
+    "google-auth>=1.19.0,<3.0.0.dev0",
     "google-auth-httplib2>=0.1.0",
     # NOTE: Maintainers, please do not require google-api-core>=2.x.x
     # Until this issue is closed
     # https://github.com/googleapis/google-cloud-python/issues/10566
-    "google-api-core >= 1.31.5, <3.0.0dev,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.0",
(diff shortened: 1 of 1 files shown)


con-f-use:
I signed your agreement, please fix the commit message yourself.

PR #2166, opened 2023-06-20
chore: use branch from fork in github action

parthea:


shakaran:
@parthea it seems that you make this PR in official branch when you want this only in your fork branch. I will recommend you close this PR or some maintainer could close it

PR #2372, opened 2024-03-27
chore: add universe compatibility with older versions of api-core

ohmayr:


tests/test_discovery.py
```diff
@@ -2665,6 +2678,66 @@ def test_universe_env_var_configured_with_client_options_universe(self):
 
             assert tasks._universe_domain == fake_universe
 
+    else:
+        if hasattr(google.api_core.client_options.ClientOptions, "universe_domain"):
+
+            def test_client_options_universe_with_older_version_of_api_core(self):
```
parthea, line 2684: Should we also add a test case for http credentials. eg. `test_http_credentials_client_options_universe_with_older_version_of_api_core`?

parthea (approved):

tests/test_discovery.py
ohmayr: setting `clientOptions.universe_domain` will raise an exception before we even check for credentials in `build_from_document`. Therefore, testing different type of credentials while also passing in `clientOptions.universe_domain` might not add much value here.

PR #2443, opened 2024-07-17
build: fetch discovery artifacts from discovery-artifact-manager

### LINKED ISSUE #2444: Remove `FLAGS.discovery_uri_template` from function `document_api`
See review feedback from PR https://github.com/googleapis/google-api-python-client/pull/2443 where `FLAGS` shouldn't be referenced directly in functions. 

See the code below, specifically line 410 where we have `FLAGS.discovery_uri_template`
https://github.com/googleapis/google-api-python-client/blob/0cb72664d833c4c8e05021a12659f14bc916fea5/describe.py#L392-L412

parthea:
Fixes b/352084616
Fixes #2444

### DIFF
--- a/describe.py
+++ b/describe.py
@@ -408,14 +408,7 @@ def document_api(
             artifacts should be saved.
     """
     http = build_http()
-    # TODO(https://github.com/googleapis/google-api-python-client/issues/2444):
-    # We shouldn't be referencing FLAGS directly from within function
-    resp, content = http.request(
-        uri
-        or uritemplate.expand(
-            FLAGS.discovery_uri_template, {"api": name, "apiVersion": version}
-        )
-    )
+    resp, content = http.request(uri)
 
     if resp.status == 200:
         discovery = json.loads(content)
@@ -519,10 +512,13 @@ def generate_all_api_documents(
     if resp.status == 200:
         directory = json.loads(content)["items"]
         for api in directory:
+            uri = uritemplate.expand(
+                discovery_uri_template or FLAGS.discovery_uri_template or api["discoveryRestUrl"],
+                {"api": api["name"], "apiVersion": api["version"]}
+            )
             document_api(
                 api["name"],
                 api["version"],
-                discovery_uri_template or api["discoveryRestUrl"],
                 doc_destination_dir,
                 artifact_destination_dir,
             )
--- a/describe.py
+++ b/describe.py
@@ -495,7 +495,7 @@ def generate_all_api_documents(
     directory_uri=DIRECTORY_URI,
     doc_destination_dir=BASE,
     artifact_destination_dir=DISCOVERY_DOC_DIR,
-    discovery_uri_template=None,
+    discovery_uri_template=DISCOVERY_URI_TEMPLATE,
 ):
     """Retrieve discovery artifacts and fetch reference documentations
     for all apis listed in the public discovery directory.
--- a/describe.py
+++ b/describe.py
@@ -421,8 +421,8 @@ def document_api(
         # Use `uri` argument of `document_api` or
         # create the uri from `FLAGS.discovery_uri_template`
         uri = uri or uritemplate.expand(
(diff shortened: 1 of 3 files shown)


describe.py
```diff
@@ -402,14 +402,23 @@ def document_api(
...
-        or uritemplate.expand(
+    # Use the discovery_uri_template to create the uri if provided
+    if discovery_uri_template:
+        uri = uritemplate.expand(
+            discovery_uri_template, {"api": name, "apiVersion": version}
+        )
+
+    if not uri:
+        uritemplate.expand(
             FLAGS.discovery_uri_template, {"api": name, "apiVersion": version}
```
chingor13, line 417: This seems like an anti-pattern (depending on a global) that can be cleaned up in the future.

chingor13 (approved):

describe.py
parthea: The order for constructing `uri` is use `discovery_uri_template` if provided, then use `uri` argument if provided, then fallback to `FLAGS.discovery_uri_template`

It's not clear how this can be cleaned up to preserve this order

describe.py
```diff
@@ -402,14 +407,25 @@ def document_api(
             documentation should be saved.
         artifact_destination_dir (str): relative path where the discovery
             artifacts should be saved.
+        discovery_uri_template (str): URI template of the API's discovery document.
```
ohmayr, line 410: Should `discovery_uri_template` be marked as `Optional[str]`?

describe.py
ohmayr: We could instead do something like this:

```python3

uri = (discovery_uri_template and discovery_uri_template, {"api": name, "apiVersion": version}) 
            or uri or uritemplate.expand(FLAGS.discovery_uri_template, {"api": name, "apiVersion": version})
```

describe.py
```diff
@@ -503,6 +520,7 @@ def generate_all_api_documents(
...
```
ohmayr, line 523: Same comment here.

ohmayr (approved):

describe.py
```diff
@@ -516,6 +534,7 @@ def generate_all_api_documents(
                 api["discoveryRestUrl"],
                 doc_destination_dir,
                 artifact_destination_dir,
+                discovery_uri_template,
```
vchudnov-g, line 532: Consider having `document_api` not take an extra parameter:

```python
            document_api(
                api["name"],
                api["version"],
                discovery_uri_template or FLAGS.discovery_uri_template or api["discoveryRestUrl"],
                doc_destination_dir,
                artifact_destination_dir,              
            )
```

Then, in `document_api`, you can simply have
```python
uri = uri.expand(...)
```
without any conditionals.

Note also that I have `FLAGS.discovery_uri_template` in the second position; that makes more sense, I think. But I also think we shouldn't be referencing `FLAGS` directly from within the functions, but it should be passed in.


```

describe.py
vchudnov-g: See my comment below which touches on this.

parthea: Fixed in https://github.com/googleapis/google-api-python-client/pull/2443/commits/7a35f8c26b0452a150cac1e3b02a82fc6a1bc30b

describe.py
parthea: Fixed in https://github.com/googleapis/google-api-python-client/commit/3863180415eea21b2d74d4e58e65ccb83ed82cae

describe.py
vchudnov-g: Also, I'm not sure why the flag `--discovery_uri_template` still defaults to `DISCOVERY_URI`. I think we should change the latter symbol to the value of `DISCOVERY_URI_TEMPLATE` that you  have in scripts/updatediscoveryartifacts.py

ohmayr (approved):

describe.py
parthea: Fixed. Please could you take another look?

ohmayr (approved):

Issue #512, opened 2018-05-30
Memory leak in file download from google drive


alexanderwhatley:
I'm running the following code in google colab to download a large file (approximately 2.5 GB).
```
import io
from google.colab import auth
from googleapiclient.discovery import build
from googleapiclient.http import MediaIoBaseDownload
auth.authenticate_user()

drive_service = build('drive', 'v3')
file_id = '1M5ZTq0Iu0kttOZb1FpDPRG-TsagdYbIQ' 
request = drive_service.files().get_media(fileId=file_id)
downloaded = io.BytesIO()
downloader = MediaIoBaseDownload(downloaded, request)
done = False
while done is False:
  status, done = downloader.next_chunk()
  print('Percent downloaded', int(100 * status.progress()))
downloaded.seek(0)
f = open('data.tar.gz', 'wb')
f.write(downloaded.read())
f.close()
del drive_service, downloader, downloaded, f
printm()
```

I am using the following script to determine memory usage:
```
!ln -sf /opt/bin/nvidia-smi /usr/bin/nvidia-smi
!pip install gputil
!pip install psutil
!pip install humanize
import psutil
import humanize
import os
def printm():
 process = psutil.Process(os.getpid())
 print("Gen RAM Free: " + humanize.naturalsize( psutil.virtual_memory().available ), " I Proc size: " + humanize.naturalsize( process.memory_info().rss))
printm()
```
However, it appears that there is a memory leak in the api client. I print the amount of memory usage at the very beginning, before running anything:
```
Requirement already satisfied: gputil in /usr/local/lib/python3.6/dist-packages (1.3.0)
Requirement already satisfied: numpy in /usr/local/lib/python3.6/dist-packages (from gputil) (1.14.3)
Requirement already satisfied: psutil in /usr/local/lib/python3.6/dist-packages (5.4.5)
Requirement already satisfied: humanize in /usr/local/lib/python3.6/dist-packages (0.5.1)
Gen RAM Free: 13.0 GB  I Proc size: 146.8 MB
```
After running the download script, which as you can see deletes all of the variables, as well as the `%reset` magic, the amount of memory is still much less than before:
```
Requirement already satisfied: gputil in /usr/local/lib/python3.6/dist-packages (1.3.0)
Requirement already satisfied: numpy in /usr/local/lib/python3.6/dist-packages (from gputil) (1.14.3)
Requirement already satisfied: psutil in /usr/local/lib/python3.6/dist-packages (5.4.5)
[...]

Issue #198, opened 2016-03-08
Large memory footprint


dabaitu:
I've recently noticed my download script which uses chunked downloading of 250MB would use nearly 2GB of ram when downloading a file of size ~7GB. The string buffer gets cleaned for every 250MB chunk, so I don't expect that large of a ram usage.

Please let me know what other information is needed here.

mattwhisenhunt:
Thank you for your bug report. This client library is currently in maintenance mode. We are fixing necessary bugs and adding essential features to ensure this library continues to meet your needs for accessing Google APIs. After careful consideration, we have decided not to fix this bug since there is an acceptable work-around. If you disagree, please feel free explain why you believe fixing it is essential for users of this client library.

zachliu:
@mattwhisenhunt What's the work-around? I'm using v1.7.4 and still experiencing the same issue.

busunkim96:
@zachliu Please open another issue referencing this one with a minimal reproduction of the problem you are experiencing. It's best for you to open a new issue as it's easy for comments on already closed issues to be lost. 

Thanks.

zachliu:
@busunkim96 Will do :+1:. It's just that Mr. Whisenhunt left a huge cliffhanger in his post: there's a workaround, but i'm not gonna say it :joy:

PR #2453, opened 2024-07-29
 fix#1706 - Issue while downloading large files 

hexadivine:
Added MediaGenBaseDownload class to fetch chunks and to keep RAM consumption linear and test cases (TestMediaGenBaseDownload class).

Fixes : #1706, #512, #198


gkevinzheng:
@hexadivine Thanks for putting the fix together, but unfortunately this repo contains legacy versions of our GCP libraries and is in maintenance mode, so we try not to make major changes here. 

If it's at all possible, I'd recommend trying to move to the more modern gapic libraries. If that's not possible, you may be able to make a fork of this code to accomplish what you need

Issue #1637, opened 2021-12-15
Asynchronous features (async, aiohttp, etc)


Vioshim:
### Hello, good evening Google devs.

 **Is your feature request related to a problem? Please describe.**
> I've been a big fan of this API and being able to use it in python, however due to it being reliant on httplib2 and 
time.sleep (According to `googleapiclient.http.HttpRequest`).
> 
> Currently most programmers that use the API do proper threading practices in order to keep using this api without it being blocking in asynchornous context.

 **Describe the solution you'd like**
> By this reason I'd like to propose making classes that are designed for asynchronous code, those can rely on aiohttp.request which can get to be advantageous for this regard.

 **Describe alternatives you've considered**
> As stated before, the current alternative that I've been using with this library is asyncio's to_thread which helps to stop the blocking issue, but it's still not that adapted to the asynchronous context.

 **Additional context**
> Not much to add in this regard.

So that's pretty much my suggestion, have a great day.

ChynnaSyas:
I'm working on this

ChynnaSyas:
Question: 

How can the Google Python API client library be enhanced to fully support asynchronous operations, ensuring non-blocking I/O and improved performance for high-throughput applications?

Hypothesis: 

To effectively add asynchronous support to the Google Python API client library, the key design considerations and challenges must be addressed:

- Asynchronous HTTP Requests: Challenge: Refactoring synchronous HTTP request functions to asynchronous without disrupting existing functionality.
- Asynchronous Authentication: Challenge: Handling OAuth 2.0 authentication and token refresh operations asynchronously.
- Concurrency: Challenge: Efficiently managing multiple concurrent API requests and handling rate limits imposed by Google APIs.
- Error Handling: Challenge: Managing errors and retries in an asynchronous context, ensuring that the application remains resilient.
- Backward Compatibility: Challenge: Maintaining backward compatibility with existing synchronous code to avoid breaking changes for current users.
- Testing: Challenge: Ensuring the reliability and performance of the new asynchronous features through comprehensive testing.

Domush:
Allow me to attempt to answer each of these, as it's been 7 months:

**Asynchronous HTTP Requests:** new async methods can be named with Async at the end, with all sync method names remaining intact. A common approach for C# libraries.

**Asynchronous Authentication:** Simply using aiohttp with await will cover this. Not an issue.

[...]

PR #2469, opened 2024-08-26
Added Asynchronous Features To Close Issue #1637

ChynnaSyas:
We have added asynchronous features for HTTP requests, authentication, as well as completed some of the checklist below:

- [ x ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-api-python-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ x ] Ensure the tests and linter pass
- [ x ] Appropriate docs were updated (if necessary)

Fixes #<1637> 🦕

### DIFF
--- a/googleapiclient/http.py
+++ b/googleapiclient/http.py
@@ -1266,7 +1266,7 @@ class AsyncHttpRequest(object): #Team_CKL ADDED CODE
             self.headers["content-length"] = str(len(self.body))
 
         # Handle retries for server-side errors.
-        resp, content = await self._retry_request(
+        resp, content = await self._retry_request_(
             http,
             num_retries,
             "request",
@@ -1346,7 +1346,7 @@ class AsyncHttpRequest(object): #Team_CKL ADDED CODE
                 start_headers["X-Upload-Content-Length"] = size
             start_headers["content-length"] = str(self.body_size)
 
-            resp, content = await self._retry_request(
+            resp, content = await self._retry_request_(
                 http,
                 num_retries,
                 "resumable URI request",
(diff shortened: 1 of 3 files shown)


googleapiclient/http.py
```diff
@@ -1161,6 +1168,384 @@ def from_json(s, http, postproc):
...
+                self.uri,
+                method=self.method,
+                body=self.body,
+                headers=start_headers,
+            )
+
+            if resp.status == 200 and "location" in resp.headers:  #Potential Tech Issue problem
+                self.resumable_uri = resp.headers["location"] #Potential Tech Issue problem
+            else:
+                raise ResumableUploadError(resp, content)
```
Prateikx, line 1364: **✨ AI Suggestion:** Directly accessing `resp.headers["location"]` can lead to a `KeyError` if the header is unexpectedly missing, even if the status code is 200. Use the `.get()` method for safer access and raise a more specific error if the essential header is absent.
```suggestion
            location = resp.headers.get("location")
            if resp.status == 200 and location:
                self.resumable_uri = location
            else:
                # Raise error if status is 200 but location is missing, or if status is not 200
                raise ResumableUploadError(
                    resp, content, message="Missing 'location' header in resumable URI response."
                )
```

googleapiclient/http.py
```diff
@@ -1161,6 +1168,384 @@ def from_json(s, http, postproc):
...
+        elif resp.status == 308:
+            self._in_error_state = False
+            # A "308 Resume Incomplete" indicates we are not done.
+            try:
+                self.resumable_progress = int(resp["range"].split("-")[1]) + 1
+            except KeyError:
+                # If resp doesn't contain range header, resumable progress is 0
+                self.resumable_progress = 0
+            if "location" in resp:
+                self.resumable_uri = resp.headers["location"]  #Potential Tech Issue problem
```
Prateikx, line 1467: **✨ AI Suggestion:** Directly accessing `resp.headers["location"]` after a 308 response can lead to a `KeyError` if the header is missing. Use `.get()` for safer access to prevent potential runtime errors.
```suggestion
                except KeyError:
                    # If resp doesn't contain range header, resumable progress is 0
                    self.resumable_progress = 0
                # Use .get() for safer access to the location header
                location = resp.headers.get("location")
                if location:
                    self.resumable_uri = location
```

gkevinzheng:
Thanks for proposing the new feature! Unfortunately this repo contains legacy versions of our GCP libraries and is in maintenance mode, so we try not to make major changes here.

If it's at all possible, I'd recommend trying to move to the more modern gapic libraries. If that's not possible, you may be able to make a fork of this code to accomplish what you need.

PR #2486, opened 2024-09-18
Remove "apiclient" package

### LINKED ISSUE #2485: Package writes to "apiclient" directory when installed
#### Environment details

  - OS type and version: Linux
  - Python version: 3.12.5
  - pip version: 24.2
  - `google-api-python-client` version:  2.145.0

#### Steps to reproduce

  1. `python -m virtualenv venv`
  2. `. venv/bin/activate`
  3. `pip install api-client`
  4. `python -c 'from apiclient import HeaderAuthentication'`
  5. `pip install firebase-admin`
  6. `python -c 'from apiclient import HeaderAuthentication'`

I expect step 4 to be the same as step 2 here, but it results in a stack trace since `firebase-admin` depends on `google-api-python-client` which is overwriting `.../site-packages/apiclient/`. What's even worse is that some package managers (e.g., poetry) seem to install this in a random order resulting in it sometimes working and sometimes not.

#### Code example

```python
from apiclient import HeaderAuthentication
```

#### Stack trace
```
  File "<string>", line 1, in <module>
ImportError: cannot import name 'HeaderAuthentication' from 'apiclient' (/private/tmp/test1/venv/lib/python3.11/site-packages/apiclient/__init__.py)
```

kruton:
`apiclient` package was created in 2014 as a stop-gap to help people transition from "`apiclient`" to `googleapiclient`. This transition should have been completed in the past 10 years.

However now this is creating package pollution that interferes with another PyPI package called [`api-client`](https://pypi.org/project/api-client/) that also writes to `…/site-packages/apiclient/` and conflicts with this package.

Fixes #2485 🦕


kruton:
Rebased onto newest commit.

kruton:
Rebased onto head

daniel-sanche:
From https://github.com/googleapis/google-api-python-client/issues/2485:
> Thanks for putting the fix together, but unfortunately this repo contains legacy versions of our GCP libraries and is in maintenance mode, so we try not to make major changes here. Removing the `apiclient` alias would be a breaking change, which could have a large impact on existing user code, so it's not a change we'll be able to make at this time
> 
> If it's at all possible, I'd recommend trying to move to the more modern gapic libraries ([This](https://github.com/firebase/firebase-admin-python) is the gapic equivalent of firebase admin). If that's not possible, you may be able to make a fork of this code to accomplish what you need

PR #2573, opened 2025-03-04
chore: expose Resource class through __all__

cipz:
Added Reource to __all__ to explicitly export it as part of the module's public API.


chalmerlowe:
We appreciate the effort and time. Regretfully, we are gonna close this PR, as this repo is in maintenance mode.

PR #2686, opened 2025-11-20
feat: auto-enable mTLS when supported certificates are detected

agrawalradhika-cell:
The Python SDK will use a hybrid approach for mTLS enablement:

If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable is set (either true or false or any value), the SDK will respect that setting. This is necessary for test scenarios and users who need to explicitly control mTLS behavior.
If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable is not set, the SDK will automatically enable mTLS only if it detects Managed Workload Identity (MWID) or X.509 Workforce Identity Federation (WIF) certificate sources. In other cases where the variable is not set, mTLS will remain disabled.

** This change also adds unit test with respect to this change.

### DIFF
--- a/tests/test_discovery.py
+++ b/tests/test_discovery.py
@@ -141,7 +141,7 @@ def read_datafile(filename, mode="r"):
     with open(datafile(filename), mode=mode) as f:
         return f.read()
 
-def parse_version_to_tuple(version_string: str) -> ParsedVersion:
+def parse_version_to_tuple(version_string):
     """Safely converts a semantic version string to a comparable tuple of integers.
 
     Example: "4.25.8" -> (4, 25, 8)


andyrzhao (approved):
LGTM.

tests/test_discovery.py
```diff
@@ -884,6 +896,47 @@ def test_mtls_with_provided_client_cert(
...
+    @parameterized.expand(
+        [
+            ("never", "", CONFIG_DATA_WITH_WORKLOAD , REGULAR_ENDPOINT),
+            ("auto", "", CONFIG_DATA_WITH_WORKLOAD, MTLS_ENDPOINT),
+            ("always", "", CONFIG_DATA_WITH_WORKLOAD, MTLS_ENDPOINT),
+            ("never", "", CONFIG_DATA_WITHOUT_WORKLOAD, REGULAR_ENDPOINT),
+            ("auto", "", CONFIG_DATA_WITHOUT_WORKLOAD, REGULAR_ENDPOINT),
+            ("always", "", CONFIG_DATA_WITHOUT_WORKLOAD, MTLS_ENDPOINT),
+        ]
+    )
```
daniel-sanche, line 915: nit: why have the second parameter if it's always empty?

tests/test_discovery.py
```diff
@@ -884,6 +896,47 @@ def test_mtls_with_provided_client_cert(
...
+    def test_mtls_with_provided_client_cert_unset_environment_variable(
+        self, use_mtls_env, use_client_cert, config_data, base_url
+    ):
+        if not hasattr(google.auth.transport.mtls, "should_use_client_cert"):
+            self.skipTest(
+                "The should_use_client_cert function is not available in this "
+                "version of google-auth."
+            )
```
daniel-sanche: I'd prefer if we could do a version check for this instead. And using a decorator would be better if possible

tests/test_discovery.py
```diff
@@ -961,6 +1014,61 @@ def test_mtls_with_default_client_cert(
...
+                "os.environ", {"GOOGLE_API_USE_CLIENT_CERTIFICATE": use_client_cert}
+            ):
+                with mock.patch("builtins.open", m):
+                    with mock.patch.dict("os.environ", {"GOOGLE_API_CERTIFICATE_CONFIG": config_filename}):
+                        plus = build_from_document(
+                        discovery,
+                        credentials=self.MOCK_CREDENTIALS,
+                        adc_cert_path=self.ADC_CERT_PATH,
+                        adc_key_path=self.ADC_KEY_PATH,
+                        )
```
daniel-sanche: the indentation here looks off

tests/test_discovery.py
```diff
@@ -961,6 +1014,61 @@ def test_mtls_with_default_client_cert(
...
+            ("always", "", CONFIG_DATA_WITHOUT_WORKLOAD, MTLS_ENDPOINT),
+        ]
+    )
+    @mock.patch(
+        "google.auth.transport.mtls.has_default_client_cert_source", autospec=True
+    )
+    @mock.patch(
+        "google.auth.transport.mtls.default_client_encrypted_cert_source", autospec=True
+    )
+    def test_mtls_with_default_client_cert_with_unset_environment_variable(
```
daniel-sanche, line 1053: can you add docstrings explaining the difference between this and `test_mtls_with_provided_client_cert_unset_environment_variable`?

agrawalradhika-cell: Done

tests/test_discovery.py
agrawalradhika-cell: Fixed

tests/test_discovery.py
agrawalradhika-cell: Updated

tests/test_discovery.py
agrawalradhika-cell: for showcasing the unset GOOGLE_API_USE_CLIENT_CERTIFICATE flag

daniel-sanche: I think that could be shown easily in the test body instead of the parameters (and maybe even in the test name)

But just a small thought for the future, you don't need to change it

tests/test_discovery.py
```diff
@@ -884,6 +879,50 @@ def test_mtls_with_provided_client_cert(
...
+        self, use_mtls_env, use_client_cert, config_data, base_url
+    ):
+        """Tests that mTLS is correctly handled when a client certificate is provided.
+
+        This test case verifies that when a client certificate is explicitly provided
+        via `client_options` and GOOGLE_API_USE_CLIENT_CERTIFICATE is unset, the 
+        discovery document build process correctly configures the base URL for mTLS 
+        or regular endpoints based on the `GOOGLE_API_USE_MTLS_ENDPOINT` environment variable.
+        """
+        if hasattr(google.auth.transport.mtls, "should_use_client_cert"):
```
daniel-sanche, line 931: Using conditionals like this is dangerous, because we could get into a state where the tests are being skipped, but look green

I suggest:
1. Check against the auth lib version, because we always expect newer versions to support should_use_client_cert (and it should alert us if they stop!)
2. use the pytest.mark.skipif decorator around the function, to automatically skip for unsupported versions

```
from google.auth import __version__ as auth_version
from google.api_core import parse_version_to_tuple

@pytest.mark.skipif(
    parse_version_to_tuple(auth_version) < (2,43,0), 
    reason="automatic mtls only enabled in google-auth<=2.43.0"
)
 def test_mtls_with_provided_client_cert_unset_environment_variable(...):
```

agrawalradhika-cell: Updated based on your comment

daniel-sanche (dismissed):
LGTM once tests pass

daniel-sanche (approved):

PR #2731, opened 2026-03-21
fix: retry on http 400 failedPrecondition

### LINKED ISSUE #2723: HTTP 400 / preconditionFailed should be retried
#### Environment details

  - OS type and version: MacOS 26
  - Python version: Python 3.12.11
  - pip version: uv
  - `google-api-python-client` version: n/a

#### Steps to reproduce

Intermittent HTTP 400 / preconditionFailed on Gmail API.  A casual web search suggests that other Google APIs also issue preconditionFailed.  At least in my use case, retrying with exponential backoff always works.  I cannot assert that in all use cases that this is the correct behavior.

#### Code example

N/A

codeXsidd:
Screenshot 2026-03-21 100631
Resolves #2723


daniel-sanche (approved):
LGTM

ohmayr (approved):

Issue #2755, opened 2026-05-31
test: strict positional parameter enforcement is ignored under pytest


Capstan:
#### Environment details

  - OS type and version: Linux (gLinux / Debian based)
  - Python version: `3.13.12` (also affects all supported Python versions `3.10+`)
  - pip version: `26.0.1`
  - `google-api-python-client` version: `2.197.0` (at commit `6e471c075039dfef24e28d11658e03d5c949c7c3`)

#### Steps to reproduce

  1. Check out the `main` branch of `google-api-python-client`.
  2. Set up a clean virtual environment and install test dependencies (`pytest`, `mock`, `parameterized`).
  3. Run the strict positional parameters enforcement test directly using `pytest`:
     ```bash
     pytest tests/test_discovery.py -k test_tests_should_be_run_with_strict_positional_enforcement
     ```

#### Code example

The bug lies in how strict argument enforcement is configured for the test suite in `tests/__init__.py`. 

Historically, the enforcement was configured inside the nose-legacy `setup_package()` hook:
```python
# tests/__init__.py
from googleapiclient import _helpers as util

def setup_package():
    """Run on testing package."""
    util.positional_parameters_enforcement = "EXCEPTION"
```

Because modern test runs in this repository (including via `nox` sessions) run under **`pytest`**, the `setup_package()` package-level hook is ignored. As a result, the tests run in warning-only mode, causing strict parameter assertion tests to be bypassed and fail with an `AssertionError` instead of raising `TypeError`.

Since the repository has migrated to `pytest`, this legacy hook is dead code. We should configure the variable directly at the package module level and completely remove the unused `setup_package` hook:

```diff
# tests/__init__.py
from googleapiclient import _helpers as util

+util.positional_parameters_enforcement = "EXCEPTION"
-
-def setup_package():
-    """Run on testing package."""
-    util.positional_parameters_enforcement = "EXCEPTION"
```

#### Stack trace

When running the tests using `pytest`, the test fails with the following output:

```
=================================== FAILURES ===================================
_ DiscoveryErrors.test_tests_should_be_run_with_strict_positional_enforcement __

self = <tests.test_discovery.DiscoveryErrors testMethod=test_tests_should_be_run_with_strict_positional_enforcement>

    def test_tests_should_be_run_with_strict_positional_enforcement(self):
        try:
            plus = build("plus", "v1", None, static_discovery=False)
>           self.fail("should have raised a TypeError exception over missing http=.")
[...]

PR #2801, opened 2026-09-01
fix: restore shadowed test_discovery_http_is_closed test

### LINKED ISSUE #2757: test: connection leak test test_discovery_http_is_closed is shadowed and syntactically invalid
### **Environment details**

  - OS type and version: Linux (gLinux / Debian based)
  - Python version: `3.13.12` (also affects all supported Python versions `3.10+`)
  - pip version: `26.0.1`
  - `google-api-python-client` version: `2.197.0` (at commit `6e471c075039dfef24e28d11658e03d5c949c7c3`)

---

### **Steps to reproduce**

1. Look at `tests/test_discovery.py` on `main`. Note that there are **two separate definitions** for `class Discovery(unittest.TestCase)` (one at line 498 and one at line 1566).
2. In Python, defining a class twice in the same module causes the second definition to silently overwrite the first. This means the first `Discovery` class—which contains the test case **`test_discovery_http_is_closed`**—is completely dead code and is never executed by `pytest`.
3. Attempting to manually extract and execute this test case crashes immediately because `HttpMock.close()` is a regular function, but the test tries to call `assert_called_once()` on it.

---

### **Code example**

The shadowed class in `tests/test_discovery.py` contains this test definition:
```python
class Discovery(unittest.TestCase):
    def test_discovery_http_is_closed(self):
        http = HttpMock(datafile("malformed.json"), {"status": "200"})
        # build() is called without passing the 'http' mock client!
        service = build("plus", "v1", credentials=mock.sentinel.credentials)
        # Fails: HttpMock.close is a standard method, not a mock object
        http.close.assert_called_once()
```

#### **Historical Context**
This bug was introduced in **September 2020** by commit **`98888dadf`** ("fix: add method to close httplib2 connections (#1038)"). 
That commit added `test_discovery_http_is_closed` by declaring a new `class Discovery(unittest.TestCase)` block near the top of the file, without realizing that a `class Discovery(unittest.TestCase)` block was already defined and active near the bottom of the file (which immediately shadowed the new test).

#### **The Proposed Solution**
1. Delete the duplicate class definition at line 498.
2. Restore `test_discovery_http_is_closed` inside the active `Discovery` class (at line 1566).
3. Correct the test case to use standard `unittest.mock.patch` on `httplib2.Http` to verify that the temporary HTTP client created by `build()` to fetch the discovery document is safely closed after initialization completes, preventing connection/socket leaks.

```python
    @mock.patch("httplib2.Http")
    def test_discovery_http_is_closed(self, mock_http_class):
        mock_http = mock_http_class.return_value
        mock_http.request.return_value = (
            httplib2.Response({"status": "200"}),
            read_datafile("plus.json"),
        )
        service = build("plus", "v1", static_discovery=False)
        mock_http.close.assert_called_once()
```

---

### **Stack trace**

If you isolate the legacy test case and run it, it crashes immediately with an `AttributeError`:

```
AttributeError: 'function' object has no attribute 'assert_called_once'
```

skippdot:
Fixes #2757

`tests/test_discovery.py` has contained two `class Discovery(unittest.TestCase)` definitions since September 2020 (#1038). The second definition silently shadows the first, so its only test, `test_discovery_http_is_closed`, has never been collected — pytest picks up 0 instances of it on `main`. The dead test also could not have passed if it ever ran: it calls `assert_called_once()` on `HttpMock.close`, a plain method, which raises `AttributeError`.

### Changes

- Remove the shadowed duplicate class.
- Re-add `test_discovery_http_is_closed` inside the surviving `Discovery` class, rewritten to patch `httplib2.Http` so the assertion targets a real mock. It verifies that `build()` closes the temporary http client it creates to fetch the discovery document (`discovery.py`, the `discovery_http.close()` call).
- Pass `cache_discovery=False` in addition to `static_discovery=False`, so the discovery document is always fetched through the mocked client — a discovery-cache hit cannot bypass the code path under test.

One file, +14/−7.

### Verification

- Collected tests in `tests/test_discovery.py`: 173 → 174.
- Mutation check: with the `discovery_http.close()` call removed from `googleapiclient/discovery.py`, the restored test fails; with it in place, the test passes — so it genuinely guards the close behavior.
- Full `tests/` folder on Python 3.14: 339 passed, 1 skipped.
- `black --check` clean; `flake8 --select=E9,F63,F7,F82` (the set CI enforces) clean.

Note: running `tests/test_discovery.py` in isolation shows one pre-existing failure, `DiscoveryErrors::test_tests_should_be_run_with_strict_positional_enforcement` — that is #2755, reproducible on an unmodified `main`, and untouched by this PR.


skippdot:
Noting for reviewers: #2758, opened earlier, addresses the same issue (#2757) with a similar approach (remove the shadowed class, re-add the test with `httplib2.Http` patched). Differences in this PR:

- passes `cache_discovery=False` in addition to `static_discovery=False`, so a discovery-cache hit cannot bypass the mocked client and make the test fail spuriously;
- patches via `http_class.return_value` instead of a two-mock `side_effect`, so the test does not depend on `build()` constructing exactly two `Http` instances in a particular order;
- mutation-checked: with the `discovery_http.close()` call removed from `discovery.py`, the test fails.

#2758 additionally asserts that the caller-facing http client is *not* closed, which is a nice extra check. Happy to fold that assertion in here, or to defer to #2758 if maintainers prefer it — flagging the overlap so both are not reviewed in parallel unknowingly.

PR #2804, opened 2026-09-01
fix: import urllib.parse explicitly instead of relying on side effects

### LINKED ISSUE #2803: googleapiclient modules use urllib.parse but only import urllib, crashing when urllib.parse was not imported elsewhere
#### Environment details

- OS type and version: macOS 15 (also reproduced on Linux)
- Python version: 3.14.6 (affects all supported versions, 3.7+ semantics are identical)
- pip version: 25.x
- `google-api-python-client` version: `main` at 65bbea25 (release 2.200.0)

#### Steps to reproduce

Run this two-line program (no other imports):

```python
import googleapiclient._helpers as h
h._add_query_parameter("http://example.com", "a", "b")
```

Result:

```
  File ".../googleapiclient/_helpers.py", line 183, in _update_query_params
    parts = urllib.parse.urlparse(uri)
            ^^^^^^^^^^^^
AttributeError: module 'urllib' has no attribute 'parse'
```

#### Root cause

Four modules import the bare `urllib` package but call `urllib.parse.*` functions:

| module | `urllib.parse` call sites |
|---|---|
| `googleapiclient/_helpers.py` | 4 |
| `googleapiclient/discovery.py` | 6 |
| `googleapiclient/http.py` | 6 |
| `googleapiclient/model.py` | 1 |

In Python, `import urllib` does **not** import its submodules; `urllib.parse` only becomes accessible as an attribute of the package after some module somewhere executes `import urllib.parse` (or `from urllib.parse import ...`). The library currently works only because `httplib2`, imported transitively by `discovery.py`/`http.py`/`model.py`, happens to register `urllib.parse` first. `_helpers.py` imports nothing but the standard library, so using it before anything has imported `urllib.parse` crashes as above.

#### Visible symptom in this repository's own test suite

```
pytest tests/test__helpers.py
# 7 failed, 3 passed — every failure is AttributeError: module 'urllib' has no attribute 'parse'
```

The same tests pass in a full-suite run because other test modules import `urllib.parse` first — the same accidental-transitive-import mechanism.

#### Proposed fix

Replace `import urllib` with `import urllib.parse` in the four modules listed above. Call sites keep their current `urllib.parse.urlparse(...)` spelling, so the change is one line per module. This removes the undeclared dependency on third-party import side effects for all four modules, not just the one that currently crashes.

This is the fix: https://github.com/googleapis/google-api-python-client/pull/2804

skippdot:
Fixes #2803

Four modules import the bare `urllib` package but call `urllib.parse.*`: `_helpers.py` (4 call sites), `discovery.py` (6), `http.py` (6), `model.py` (1). `import urllib` does not import submodules — `urllib.parse` only becomes an attribute of the package after some module executes `import urllib.parse`. Today that happens as a side effect of `httplib2` being imported transitively, which is why the bug is invisible in most runs.

`_helpers.py` has no third-party imports, so it crashes outright:

```python
import googleapiclient._helpers as h
h._add_query_parameter("http://example.com", "a", "b")
# AttributeError: module 'urllib' has no attribute 'parse'
```

The same mechanism is why `pytest tests/test__helpers.py` in isolation fails 7 of 10 tests on `main` while the full-suite run passes.

### Change

Replace `import urllib` with `import urllib.parse` in the four modules. Call sites are untouched — `urllib.parse.urlparse(...)` keeps its current spelling — so the diff is one line per module, +4/−4. The other three modules are included because they carry the same undeclared dependency on `httplib2`'s import side effects, just without a currently-reachable crash.

### Verification (Python 3.10 and 3.14)

- The two-line reproducer above: crashed before, returns `http://example.com?a=b` after.
- Fresh-subprocess import-and-use probe of each of the four modules: all OK.
- `pytest tests/test__helpers.py` in isolation: 7 failed / 3 passed → **10 passed**.
- Full `tests/` folder: 338 passed, 1 skipped on both Python versions.
- `black --check` clean; `flake8 --select=E9,F63,F7,F82` (the set CI enforces) clean.

Closes https://github.com/googleapis/google-api-python-client/pull/2807


daniel-sanche (approved):
LGTM

parthea (approved):

